PR #43, opened 2023-06-01
Make presentation mode more handy by arrow keys to change slides

ahmadkadri:

Screenshot 2023-06-01 145404


src/components/common/Toolbar/Toolbar.tsx
```diff
@@ -62,53 +75,62 @@ export function Toolbar({ children, ...props }: ToolbarProps) {
     [focusedToolbarKey]
   );
 
-  const handleKeyDown = useCallback((event: KeyboardEvent) => {
-    const list = ref.current;
-    const currentFocus = document.activeElement;
+  const handleKeyDown = useCallback(
```
dhenneke, line 78: There are some issues when I compare this to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/toolbar_role:

1. A vertical toolbar should use the up/down buttons but not the left/right buttons.
2. A vertical toolbar has `aria-orientation="vertical"` set. (This can also be used for the styling).

We should also have the same tests for the vertical toolbar that we also already have for the horizontal toolbar.

src/components/PresentBar/PresentBar.tsx
```diff
@@ -14,26 +14,76 @@
...
+  const isViewingPresentation = state.type === 'presenting';
+
+  const handleToNextSlideClick = useCallback(() => {
+    if (activeSlideId) {
+      const activeSlideIndex = findIndex(
+        slideIds,
+        (value) => value === activeSlideId
+      );
+      whiteboardInstance.setActiveSlideId(slideIds[activeSlideIndex + 1]);
+    }
```
dhenneke, line 56: You can also use `indexOf` instead of lodash.

```suggestion
    if (activeSlideId) {
      const activeSlideIndex = slideIds.indexOf(activeSlideId);
      whiteboardInstance.setActiveSlideId(slideIds[activeSlideIndex + 1]);
    }
```

src/components/common/Toolbar/Toolbar.tsx
```diff
@@ -117,6 +139,7 @@ export function Toolbar({ children, ...props }: ToolbarProps) {
       onKeyDown={handleKeyDown}
       disableGutters
       variant="dense"
+      toolbarDirection={toolbarDirection}
```
dhenneke, line 142: The tests show the following warning:

> Warning: React does not recognize the `toolbarDirection` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `toolbardirection` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

There are different options to solve that. I think the easiest one is to use the `aria-orientation` to decide the `flexDirection` of the component.

src/components/common/Toolbar/Toolbar.tsx
```diff
@@ -62,53 +75,62 @@ export function Toolbar({ children, ...props }: ToolbarProps) {
...
+          // Up and down arrow keys are used to navigate in a single radio group in a toolbar
+          // https://developer.mozilla.org/en-US/docs/web/accessibility/aria/roles/radiogroup_role#keyboard_interactions
+          if (isRadioInput(currentFocus)) {
+            event.preventDefault();
+            moveFocus(
+              findParentRadioGroup(currentFocus, list),
+              currentFocus,
+              previousItem
+            );
+          } else if (toolbarDirection && toolbarDirection === 'column') {
```
dhenneke, line 104:
```suggestion
          } else if (toolbarDirection === 'column') {
```

If `toolbarDirection` is `undefined`, it can't be `"column"`. You also set the default value to `"row"` so it will never be `undefined`.

src/components/common/Toolbar/Toolbar.tsx
```diff
@@ -32,21 +32,34 @@ import {
...
 
+export type ToolbarDirection = 'row' | 'column';
+
 export type ToolbarProps = PropsWithChildren<
-  Partial<React.ComponentPropsWithRef<'div'> & { sx?: SxProps<Theme> }>
+  Partial<
+    React.ComponentPropsWithRef<'div'> & { sx?: SxProps<Theme> } & {
+      toolbarDirection?: ToolbarDirection;
+    }
+  >
```
dhenneke, line 55: The `Partial<>` was wrong before. This can be simplified to:

```suggestion
  Partial<React.ComponentPropsWithRef<'div'>> & {
    sx?: SxProps<Theme>;
    toolbarDirection?: ToolbarDirection;
  }
```

src/components/PresentBar/PresentBar.tsx
```diff
@@ -43,6 +93,24 @@ export function PresentBar() {
...
         />
       )}
+      {isViewingPresentation && (
+        <>
+          <ToolbarButton
+            aria-label={presentToolsBarPreviousSlide}
+            disabled={isFirstSlideActive}
+            onClick={handleToPreviousSlideClick}
+          >
+            <ChevronLeftIcon />
```
dhenneke, line 103: The figma design uses `ArrowBackIosIcon` and `ArrowForwardIosIcon`. Is there a reason why it was changed to `Chevron*Icon`?

src/components/common/Toolbar/Toolbar.tsx
ahmadkadri: i already did the lowercase solution

dhenneke: Please don't. That is not the solution we want to use!

nils-nord:
Test passed. 
Tested using Mozilla Firefox & Google Chrome in latest public releases

src/components/PresentBar/PresentBar.tsx
```diff
@@ -43,6 +86,24 @@ export function PresentBar() {
           checkedIcon={<PresentToAllIcon />}
         />
       )}
+      {isViewingPresentation && (
```
dhenneke, line 89: The name of the variable is wrong because it is only displayed for the presenter not for the ones that view the presentation. But this file uses the pattern to use the state.type variable directly so we can change it to:

```suggestion
      {state.type === 'presenting' && (
```

and skip the variable.

src/components/PresentBar/PresentBar.tsx
```diff
@@ -14,26 +14,69 @@
...
+  const whiteboardInstance = useActiveWhiteboardInstance();
+  const { activeSlideId, isFirstSlideActive, isLastSlideActive } =
+    useActiveSlide();
   const { state, togglePresentation } = usePresentationMode();
 
+  const isViewingPresentation = state.type === 'presenting';
+
+  const handleToNextSlideClick = useCallback(() => {
+    if (activeSlideId) {
+      const activeSlideIndex = slideIds.indexOf(activeSlideId);
```
dhenneke, line 50: You can also get the slideIds from the `whiteboardInstance`. Then we don't need another hook call that potentially triggers a rendering:

```suggestion
      const slideIds = whiteboardInstance.getSlideIds();
      const activeSlideIndex = slideIds.indexOf(activeSlideId);
```

.changeset/four-actors-return.md
```diff
@@ -0,0 +1,5 @@
+---
+'@nordeck/matrix-neoboard-widget': minor
+---
+
+Add slides navigation button forward and backwards to the presentBar
```
dhenneke, line 5:
```suggestion
Add buttons to navigate to the next or previous slide in a presentation.
```

dhenneke (commented):
Looks good. Only some small comments then this should be ready.

dhenneke (approved):

PR #53, opened 2023-06-07
Support editing in presentation mode

charlynguyen:

01-presentation-mode-editing-disabled
02-presentation-mode-editing-enabled

### DIFF
--- a/src/components/PresentBar/PresentBar.test.tsx
+++ b/src/components/PresentBar/PresentBar.test.tsx
@@ -252,4 +252,36 @@ describe('<PresentBar/>', () => {
       })
     ).toBeDisabled();
   });
+
+  it('should toggle the edit mode', async () => {
+    whiteboardManager
+      .getActiveWhiteboardInstance()
+      ?.getPresentationManager()
+      .startPresentation();
+
+    render(<PresentBar />, { wrapper: Wrapper });
+
+    const toolbar = screen.getByRole('toolbar', { name: 'Present' });
+
+    await userEvent.click(
+      within(toolbar).getByRole('checkbox', {
+        name: 'Enable editing',
+        checked: false,
+      })
+    );
+
+    await userEvent.click(
+      within(toolbar).getByRole('checkbox', {
+        name: 'Disable editing',
+        checked: true,
+      })
+    );
+
+    expect(
+      within(toolbar).getByRole('checkbox', {
+        name: 'Enable editing',
+        checked: false,
+      })
+    ).toBeInTheDocument();
+  });
 });
--- a/src/state/communication/messages/presentSlide.test.ts
+++ b/src/state/communication/messages/presentSlide.test.ts
@@ -33,7 +33,7 @@ describe('isValidPresentSlideMessage', () => {
       isValidPresentSlideMessage({
         senderSessionId: 'sender-session-id',
         senderUserId: '@sender-user-id',
-        content: { view: { slideId: 'slide-0' } },
+        content: { view: { isEditMode: false, slideId: 'slide-0' } },
         type: 'net.nordeck.whiteboard.present_slide',
       })
     ).toBe(true);
@@ -45,7 +45,7 @@ describe('isValidPresentSlideMessage', () => {
         senderSessionId: 'sender-session-id',
         senderUserId: '@sender-user-id',
         content: {
-          view: { slideId: 'slide-0', additional: 'tmp' },
+          view: { isEditMode: false, slideId: 'slide-0', additional: 'tmp' },
           additional: 'tmp',
         },
         type: 'net.nordeck.whiteboard.present_slide',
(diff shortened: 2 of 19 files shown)


src/components/PresentBar/PresentBar.test.tsx
```diff
@@ -252,4 +252,50 @@ describe('<PresentBar/>', () => {
...
+    render(<PresentBar />, { wrapper: Wrapper });
+
+    const toolbar = screen.getByRole('toolbar', { name: 'Present' });
+
+    expect(
+      within(toolbar).getByRole('checkbox', {
+        name: 'Enable editing',
+        checked: false,
+      })
+    ).toBeInTheDocument();
```
nurjinjafar, line 271: do we really have to check explicitly if the checkbox is in tzhe document, wouldnt it be sufficiant to directly wait for the user to click?!

ahmadkadri (dismissed):

src/state/communication/messages/presentSlide.test.ts
```diff
@@ -57,10 +57,11 @@ describe('isValidPresentSlideMessage', () => {
...
     { view: {} },
-    { view: { slideId: undefined } },
-    { view: { slideId: null } },
-    { view: { slideId: 111 } },
-    { view: { slideId: '' } },
+    { view: { isEditMode: false, slideId: undefined } },
+    { view: { isEditMode: false, slideId: null } },
+    { view: { isEditMode: false, slideId: 111 } },
+    { view: { isEditMode: false, slideId: '' } },
+    { view: { isEditMode: undefined, slideId: 'slide-0' } },
```
dhenneke, line 64: Please also check if `null` or a non-boolean like a number is rejected.

It seems tedious to test all these cases, but we found a lot of bugs in the validation schemas with these kinds of tests.

src/state/communication/messages/presentSlide.ts
```diff
@@ -21,11 +21,12 @@ import { Message } from '../connection';
 export const PRESENT_SLIDE_MESSAGE = 'net.nordeck.whiteboard.present_slide';
 
 export type PresentSlide = {
-  view: { slideId: string } | undefined;
+  view: { isEditMode: boolean; slideId: string } | undefined;
```
dhenneke, line 24: `isEditMode` could also be optional. It would mean you don't need to change so many tests and it would be a backwards-compatible change. But it's fine to have it as required field.

dhenneke, line 24: Please also document the field in `docs/model/webrtc-messages.md`.

src/state/presentationManagerImpl.ts
```diff
@@ -54,6 +54,9 @@ export class PresentationManagerImpl implements PresentationManager {
 
   private activeSlidePublisher: Subscription | undefined;
 
+  private isEditMode: boolean = false;
+  private isEditModeSubject = new BehaviorSubject<boolean>(false);
```
dhenneke, line 58: No need to explicitly specify the type in this case.

```suggestion
  private isEditModeSubject = new BehaviorSubject(false);
```

src/state/presentationManagerImpl.test.ts
```diff
@@ -334,4 +340,23 @@ describe('presentationManager', () => {
...
+      'net.nordeck.whiteboard.present_slide',
+      { view: { isEditMode: true, slideId: 'slide-0' } }
+    );
+
+    expect(communicationChannel.broadcastMessage).toHaveBeenNthCalledWith(
+      2,
+      'net.nordeck.whiteboard.present_slide',
+      { view: { isEditMode: false, slideId: 'slide-0' } }
+    );
+  });
```
dhenneke, line 361: There are more cases that are implemented but not yet tested:
* should accept the edit mode from a different user
* should deactivate the edit mode when the slide changes

src/state/presentationManagerImpl.test.ts
```diff
@@ -334,4 +340,23 @@ describe('presentationManager', () => {
 
     expect(await states).toEqual([{ type: 'idle' }]);
   });
+
+  it('should toggle the edit mode', () => {
+    presentationManager.toggleEditMode();
+    presentationManager.toggleEditMode();
+
```
dhenneke, line 347: Please also check if `observePresentationState` emits the expected state in this situation. See "should start the presentation" for an example.

src/state/communication/messages/presentSlide.ts
charlynguyen: I am going to keep it required.

src/state/presentationManagerImpl.ts
```diff
@@ -93,9 +96,12 @@ export class PresentationManagerImpl implements PresentationManager {
       .subscribe(({ senderSessionId, content }) => {
         if (content.view) {
           this.setCurrentPresenter(senderSessionId);
+          this.setEditMode(content.view.isEditMode);
+          content.view.isEditMode && whiteboardInstance.clearUndoManager();
```
dhenneke, line 100:
```suggestion
          if (content.view.isEditMode) {
            whiteboardInstance.clearUndoManager();
          }
```

dhenneke: Note that this message can be received multiple times because it is broadcasted whenever another user opens or closes the widget. This could unintuitively reset the undo stack while the edit mode is on. Maybe we could only execute it when either the slideid changes or when the editmode is enabled.

dhenneke (approved):

PR #325, opened 2024-02-08
Implement deletion of multiple selected elements

weeman1337:
output

### DIFF
--- a/src/components/ElementBar/ElementBar.test.tsx
+++ b/src/components/ElementBar/ElementBar.test.tsx
@@ -22,6 +22,7 @@ import { ComponentType, PropsWithChildren } from 'react';
 import userEvent from '@testing-library/user-event';
 import {
   mockEllipseElement,
+  mockLineElement,
   mockWhiteboardManager,
   WhiteboardTestingContextProvider,
 } from '../../lib/testUtils/documentTestUtils';
@@ -42,7 +43,16 @@ describe('<ElementBar/>', () => {
 
   beforeEach(() => {
     ({ whiteboardManager } = mockWhiteboardManager({
-      slides: [['slide-0', [['element-1', mockEllipseElement()]]]],
+      slides: [
+        [
+          'slide-0',
+          [
+            ['element-1', mockEllipseElement()],
+            ['element-2', mockLineElement()],
+            ['element-3', mockEllipseElement()],
+          ],
+        ],
+      ],
     }));
 
     activeWhiteboardInstance = whiteboardManager.getActiveWhiteboardInstance()!;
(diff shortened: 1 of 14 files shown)


src/components/ElementBar/ElementBar.test.tsx
```diff
@@ -87,16 +97,25 @@ describe('<ElementBar/>', () => {
     ).toBeInTheDocument();
   });
 
-  it('should delete an element', async () => {
+  it.each([
+    ['should delete a single element', ['element-1']],
+    ['should delete multiple elements', ['element-1', 'element-2']],
+  ])('%s', async (_testname, elementIds) => {
     const activeSlide = activeWhiteboardInstance.getSlide('slide-0');
+    activeSlide.setActiveElementIds(elementIds);
```
maheichyk, line 105: It may be good to split it into several test cases, then expect part could be more explicit (of course some expressions will be duplicated). This is how it is done currently across the tests, but it is optional.

src/state/types.ts
```diff
@@ -138,6 +138,8 @@ export type WhiteboardSlideInstance = {
   addElements(elements: Array<Element>): string[];
   /** Remove the element */
   removeElement(elementId: string): void;
+  /** Remove the elements by their IDs */
+  removeElements(elementIds: string[]): void;
```
maheichyk, line 142: I think a singular version: `removeElement` is only used in command+X shortcut currently. I don't think that we have to change the shortcut in this PR, need to do it later, but just a note to have in mind.

src/components/Shortcuts/DeleteShortcut/DeleteShortcut.test.tsx
```diff
@@ -64,20 +66,30 @@ describe('<DeleteShortcut>', () => {
...
-  it.each(['{delete}', '{backspace}'])(
-    'should delete an element with %s key',
-    async (key) => {
-      const activeSlide = activeWhiteboardInstance.getSlide('slide-0');
-      activeSlide.setActiveElementId('element-1');
+  it.each([
+    ['a single element', '{delete}', ['element-1']],
+    ['multiple elements', '{delete}', ['element-1', 'element-2']],
+    ['a single element', '{backspace}', ['element-1']],
+    ['multiple elements', '{backspace}', ['element-1', 'element-2']],
```
maheichyk, line 73: Maybe could split to several cases.

maheichyk (approved):
Looks good, thanks! Just some small optional changes/notes.

src/state/types.ts
weeman1337: You mean remove the singular version once all places are migrated to the other one?

maheichyk: yes, correct, I don't think we need singular version of delete and some others at the end.

src/components/ElementBar/ElementBar.test.tsx
weeman1337: As per discussion I want to prevent DRY here. Let's see how this is going.

src/components/Shortcuts/DeleteShortcut/DeleteShortcut.test.tsx
weeman1337: As per discussion I want to apply DRY here. Let's see how this is going.

maheichyk (approved):
Looks good!

PR #337, opened 2024-02-19
Drag selection of elements

weeman1337:
:rabbit2: performance of the drag selection itself looks good.
We have probably issues with selection when there are many elements anyway.

See the attached video.

This should be improved in a separate task.

drag-select


package.json
```diff
@@ -76,6 +76,7 @@
     "husky": "^9.0.11",
     "i18next-parser": "^8.12.0",
     "jest-axe": "^8.0.0",
+    "jest-mock": "^29.7.0",
```
weeman1337, line 79: This was installed as a sub-dependency anyway.

src/components/Layout/useLayoutState.tsx
```diff
@@ -56,6 +57,12 @@ type LayoutState = {
    * @param fullscreen - true to go fullscreen mode; false to go to non-fullscreen mode
    */
   setFullscreenMode: (fullscreen: boolean) => void;
+  /**
+   * Holds the Point where a drag-select action has been started.
+   * undefined if there is no drag-select action.
+   */
+  dragSelectStartCoords?: Point;
+  setDragSelectStartCoords: (point?: Point) => void;
```
mgcm, line 65: I understand that most of this `LayoutState` type has no code comments for other properties and setters. The change you introduced for the `fullScreenMode` feature added comments for the property and setter individually and now this change bundles both in the same comment.

Looking at this piece of code, I am seeing 3 distinct approaches and styles to comments, so I think we could settle on a single and consistent approach (even if it is not having any comments).

src/components/Whiteboard/ElementBehaviors/Selection/DragSelect.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+    act(() => {
+      setDragSelectStartCoords({ x: 0, y: 0 });
+    });
+
+    fireEvent.mouseMove(screen.getByTestId('drag-select-layer'), {
+      clientX: 50,
+      clientY: 50,
+    });
+
+    expect(activeSlide.getActiveElementIds()).toEqual(['element-0']);
```
mgcm, line 121: Would it be valuable to extend this test to a scenario where you have more than one active element? thinking of when your draggable area intersects with more than one shape / element.

src/components/Whiteboard/ElementBehaviors/Selection/calculateIntersect.ts
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @param selection - selection element
+ * @param elements - elements to test for intersection
+ * @returns IDs of the intersecting elements
+ */
+export function calculateIntersect(
+  selection: Element,
+  elements: Elements,
+): string[] {
+  const selectedElementIds = [];
```
mgcm, line 31: I think we should be consistent with the codebase and be specific about the string array type here.

src/components/Whiteboard/WhiteboardHost.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+    act(() => {
+      setDragSelectStartCoords({ x: 0, y: 0 });
+    });
+
+    expect(
+      screen.queryByRole('button', { name: 'Delete element' }),
+    ).not.toBeInTheDocument();
+  });
+
+  it('shout not show the drag select layer if the drag selection start coordinates are not set', () => {
```
mgcm, line 100: s/shout/should

src/components/Whiteboard/WhiteboardHost.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+    render(<WhiteboardHost />, { wrapper: Wrapper });
+
+    expect(screen.queryByTestId('drag-select-layer')).not.toBeInTheDocument();
+  });
+
+  it('shout show the drag select layer if the drag selection start coordinates are set', () => {
```
mgcm, line 106: s/shout/should

src/components/Whiteboard/WhiteboardHost.test.tsx
weeman1337: Why shouldn't a test shout :wink:

src/components/Layout/useLayoutState.tsx
weeman1337: I would say where it is valuable. In my opinion comments like „Set the drag select start coords“ don't make sense…

Anyway will see if I can improve here.

weeman1337: Added a comment for `setDragSelectStartCoords` as well. 

In my opinion the rule should be „where it makes sense“. Otherwise you will end up with comments like „Set drag select coords“ :laughing:

src/components/Whiteboard/ElementBehaviors/Selection/DragSelect.test.tsx
weeman1337: All this should hopefully be covered here https://github.com/nordeck/matrix-neoboard/pull/337/files#diff-141cbf23c431d9177f59fc54be3f68eabf0b58dea8bd24701892dc942be63890

But a quick test here can't hurt.

weeman1337: I've also added an element outside of the selection.

mgcm (approved):
looks good!

PR #352, opened 2024-02-29
Fix issues when adding and removing slides

weeman1337:
**Notes about the solution:**

The issue was, that removed slides are removed from the model before publishing the slide ID changes (see [here](https://github.com/nordeck/matrix-neoboard/blob/c7696799ef6bcdf07b9c2c39e7426e15dba795e9/src/state/whiteboardInstanceImpl.ts#L153)). If the components then try to access a slide whose ID still exists, it crashes.

This PR changes it to be done in two steps:

- First update the slide IDs
- Then actually remove the slides from the model

Take 2: Without delay

**Bug description:**

Precondition: 

    User A create at least two additional slides

    User B is on the same board

    both users have the slide overview open

Steps to reproduce: 

    user A delete two slides

### DIFF
--- a/.changeset/fresh-snails-hammer.md
+++ b/.changeset/fresh-snails-hammer.md
@@ -1,0 +1,5 @@
+---
+'@nordeck/matrix-neoboard-widget': patch
+---
+
+Successive deletion of slides no longer causes the remote board to crash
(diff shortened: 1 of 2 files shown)


weeman1337:
Setting back to draft for further investigation

maheichyk (approved):

PR #355, opened 2024-03-05
Support resizing of lines

charlynguyen:


src/components/Whiteboard/ElementBehaviors/Resizable/ResizeElement.test.tsx
```diff
@@ -157,4 +168,57 @@ describe('<ResizeElement />', () => {
       type: 'path',
     });
   });
+
+  it('should resize line elements', () => {
+    render(<ResizeElement elementId="element-1" />, {
+      wrapper: Wrapper,
+    });
+
+    // drag a resize box from 0,0 to 50,50
```
weeman1337, line 177: `// drag a resize box from 0,0 to 50,50` does in my opinion not fit in this case.
This test should drag the start of the line instead.

src/components/Whiteboard/ElementBehaviors/Resizable/ResizeElement.test.tsx
```diff
@@ -157,4 +168,57 @@ describe('<ResizeElement />', () => {
...
+      points: [
+        { x: 0, y: 0 },
+        { x: 50, y: 50 },
+      ],
+    });
+    fireEvent.mouseMove(resizeHandleBottomRight);
+    fireEvent.mouseUp(resizeHandleBottomRight);
+
+    const element = activeSlide.getElement('element-1');
+    expect(element).toEqual({
```
weeman1337, line 202: If the line starts at `0, 0` and ends at `1, 1` and the test drags from `0, 0` to `50, 50` then I would expect the result to be `1, 1`  /  `50, 50`.

weeman1337 (changes requested):
Basically I like the result, so that you can drag lines from their ends :+1: 

I found some issue. It looks like it still renders the other resize handles:

line-resize

Also it would be nice if you can walk me through your solution.

maheichyk (commented):
Seems to be some issue in rendering selection rectangle when several object selected using mouse:

https://github.com/nordeck/matrix-neoboard/assets/9369306/a037f388-a20c-42c2-880e-71dd96b6e00e

src/components/Whiteboard/ElementBehaviors/Resizable/ResizeElement.test.tsx
charlynguyen: Nice catch!

charlynguyen:
@weeman1337 @maheichyk The issues you mentioned above will be addressed in the next ticket (currently in progress).

src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
```diff
@@ -88,7 +172,8 @@ export function calculateDimensions(
   viewportHeight: number,
   forceLockAspectRatio: boolean = false,
   gridCellSize?: number,
-): Dimensions {
+  startingPoints?: Point[],
+): Dimensions & Points {
```
maheichyk, line 176: This change makes this API difficult to understand. The `startingPoints` that are available for both `kind`s `line` and `polyline` are passed for the line only and later on in this method if `startingPoints` are defined they are used to run specific method for the line calculation: `calculateLineDragDimensions` skipping the rest of `calculateDimensions`. The `& Points` of `Dimensions & Points` has '.points' that is an optional member (that is in contrast to all `Dimentions` members) that is set only for the `line` calculation, so making this method to return objects with different structure for both line and polyline. Would be good to improve this API that it is easier to understand how it works and what it returns for different types of elements that support resizing.

src/components/Whiteboard/ElementBehaviors/Selection/ElementBorder.tsx
```diff
@@ -81,26 +82,39 @@ export function ElementBorder({ elementIds, padding = 1 }: ElementBorderProps) {
...
+              <SelectionAnchor
+                x={lineRenderProperties.points.end.x}
+                y={lineRenderProperties.points.end.y}
+                borderWidth={selectionBorderWidth}
+              />
+            </>
+          ) : (
+            <>
+              <rect
+                data-testid={`${elementIds[0]}-border`}
```
maheichyk, line 109: It will not render a selection rectangle over several elements if one of selected elements is a line and this line is a first element on the slide.

maheichyk (changes requested):
Had some more look into the PR, few comments added.

src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
```diff
@@ -80,6 +81,89 @@ export function calculateDragDimension(
...
+    dimensions.elementKind !== 'line' ||
+    handlePosition === 'top' ||
+    handlePosition === 'right' ||
+    handlePosition === 'bottom' ||
+    handlePosition === 'left'
+  ) {
+    return dimensions;
+  }
+
+  const start = dimensions.points[0];
```
weeman1337, line 100: We should enable strict boolean checks. This may lead to start being undefined and weird things to happen.

If we consider a path as invalid if having less than two points, what about declaring the points in in the dimension type like this?

```
points: [Point, Point, ...Point[]];
```

src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
```diff
@@ -80,6 +81,89 @@ export function calculateDragDimension(
...
+  const end = dimensions.points[dimensions.points.length - 1];
+  const cursor = { x: dragX - dimensions.x, y: dragY - dimensions.y };
+
+  let newPoints;
+
+  switch (handlePosition) {
+    case 'topRight':
```
weeman1337, line 107: In my opinion lines should only have two handles: start and end.
Then this entire code here would be a no-brainer.

I wonder if things are getting too complicated here.
Resize of paths is already covered by polyline resize. What should other use-cases of resizing paths be? Also would then only one point change it's position?

Happy to discuss with you or the others.

weeman1337:
> @weeman1337 @maheichyk The issues you mentioned above will be addressed in the next ticket (currently in progress). That is already planned.

Can we add it then to the acceptance tests of this issue?

src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
charlynguyen: Fine by me. Just pulled out my stashed solution which I held back for certain reasons 🙂

src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
charlynguyen: I had the same idea and decided to stick to what we already have for lines.

charlynguyen:
> > @weeman1337 @maheichyk The issues you mentioned above will be addressed in the next ticket (currently in progress). That is already planned.
> 
> Can we add it then to the acceptance tests of this issue?

That's already the case. However, I just fixed both issues here now.

src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
weeman1337: Thanks, much better :slightly_smiling_face:  Does the same with less code.

weeman1337 (dismissed):
Looks good!

src/components/Whiteboard/ElementBehaviors/Resizable/ResizeHandle.tsx
```diff
@@ -21,9 +21,11 @@ import { ResizeHandlePosition } from './types';
 
 export function calculateResizeHandlePosition(
   position: ResizeHandlePosition,
-  containerWidth: number,
-  containerHeight: number,
+  containerWidth: number = 0,
+  containerHeight: number = 0,
   scale: number,
+  positionX: number = 0,
+  positionY: number = 0,
```
maheichyk, line 28: Would be good to have this explicit in API that for `start`and `end` (for lines) we use `positionX` and `positionY` only and `top`, `topRight` (for others) we use `containerWidth` and `containerHeight`. Otherwise looks not clear what should be the expected valid combinations or these parameters with these many default values without checking other code.

maheichyk (changes requested):
Some smaller change requested

weeman1337 (dismissed):

src/components/Whiteboard/ElementBehaviors/Resizable/ResizeHandle.tsx
```diff
@@ -16,102 +16,150 @@
...
+export type BaseParams = {
+  scale: number;
+};
+
+export type LineElementParams = BaseParams & {
+  elementKind: 'line';
+  handlePosition: LineElementResizeHandlePosition;
+  handlePositionX?: number;
+  handlePositionY?: number;
+};
```
maheichyk, line 36: `handlePositionX` and `handlePositionY` should not be optional, they are defined for lines

src/components/Whiteboard/ElementBehaviors/Resizable/ResizeHandle.tsx
```diff
@@ -16,102 +16,150 @@
...
+
+export type PolylineAndShapeElementsParams = BaseParams & {
+  elementKind: Exclude<ElementKind, 'line'>;
+  handlePosition: PolylineAndShapeElementsResizeHandlePosition;
+  containerWidth?: number;
+  containerHeight?: number;
+};
```
maheichyk, line 43: `containerWidth` and `containerHeight` are defined for non lines, so they should not be optional. There is no reason to have separate types for different categories of element kinds if we don't specify explicitly what parameters are defined for each kind.

In order to make them not optional you would need to refactor usages of `calculateResizeHandlePosition` up to `ResizeHandleWrapper`

maheichyk (changes requested):

src/components/Whiteboard/ElementBehaviors/Resizable/ResizeHandle.tsx
```diff
@@ -17,101 +17,127 @@
...
 
 export function calculateResizeHandlePosition(
-  position: ResizeHandlePosition,
-  containerWidth: number,
-  containerHeight: number,
-  scale: number,
+  params: { scale: number } & (
+    | LineElementProps
+    | PolylineAndShapeElementsProps
+  ),
```
maheichyk, line 26: It could be better to split `params` of elements and `scale` to have two separate parameters here: `params: LineElementProps | PolylineAndShapeElementsProps` and `scale: number` that will simplify handling of `params` in the places where `calculateResizeHandlePosition` is used. Maybe some better name then `params` should be used.

src/components/Whiteboard/ElementBehaviors/Resizable/ResizeHandle.tsx
```diff
@@ -17,101 +17,127 @@
...
-    case 'topLeft':
-      return {
-        x: -handleWidth / 2 - selectionBorderPadding,
-        y: -handleWidth / 2 - selectionBorderPadding,
-        width: handleWidth,
-        height: handleWidth,
-        cursor: 'nw-resize',
-      };
+  if (elementKind === 'line') {
+    const { handlePositionX = 0, handlePositionY = 0 } = params;
```
maheichyk, line 39: No need to assign default 0 to handlePositionX and `handlePositionY`, they are not optional already.

src/components/Whiteboard/ElementBehaviors/Resizable/ResizeHandle.tsx
```diff
@@ -17,101 +17,127 @@
...
+        return {
+          x: handlePositionX + handleWidth / -2,
+          y: handlePositionY + handleWidth / -2,
+          width: handleWidth,
+          height: handleWidth,
+          cursor: 'default',
+        };
+    }
+  } else {
+    const { handlePosition, containerWidth = 0, containerHeight = 0 } = params;
```
maheichyk, line 60: `handlePosition` is already defined here, no need to assign defaults to `containerWidth`, `containerHeight`

src/components/Whiteboard/ElementBehaviors/Resizable/ResizeHandle.tsx
```diff
@@ -124,30 +150,47 @@ export type DragEvent = {
...
 
 export type ResizeHandleProps = {
-  handlePosition: ResizeHandlePosition;
-  containerWidth: number;
-  containerHeight: number;
   onDrag?: Dispatch<DragEvent>;
   onDragStart?: Dispatch<DragEvent>;
   onDragStop?: Dispatch<DragEvent>;
-};
+} & (LineElementProps | PolylineAndShapeElementsProps);
```
maheichyk, line 156: I suggest to add a separate prop: `params` (or `resizeParams` or some better name) of the type `LineElementProps | PolylineAndShapeElementsProps`

src/components/Whiteboard/ElementBehaviors/Resizable/ResizeHandle.tsx
```diff
@@ -124,30 +150,47 @@ export type DragEvent = {
...
+    rectAttributesValues = calculateResizeHandlePosition({
+      scale,
+      elementKind,
+      handlePosition,
+      containerWidth,
+      containerHeight,
+    });
+  }
+
+  const { x, y, width, height, cursor } = rectAttributesValues;
```
maheichyk, line 193: if resize parameters are in a separate prop then this `if` condition is not needed and logic could work for elements of the same kind the same.

src/components/Whiteboard/ElementBehaviors/Resizable/ResizeHandle.tsx
```diff
@@ -209,7 +252,7 @@ export function ResizeHandle({
       scale={scale}
     >
       <rect
-        data-testid={`resize-handle-${cursor}`}
+        data-testid={`resize-handle-${rest.handlePosition}`}
```
maheichyk, line 255: please use destructuring to  create a `handlePosition` before: `const { handlePosition } = ...`

src/components/Whiteboard/ElementBehaviors/Resizable/ResizeElement.tsx
```diff
@@ -16,39 +16,41 @@
...
-  handlePosition: ResizeHandlePosition;
   startDimensions: Dimensions | undefined;
-  containerWidth: number;
-  containerHeight: number;
   forceLockAspectRatio?: boolean;
   onDrag: Dispatch<Dimensions>;
   onDragStart: Dispatch<DragEvent>;
   onDragStop: Dispatch<DragEvent>;
-};
+} & (LineElementProps | PolylineAndShapeElementsProps);
```
maheichyk, line 45: I suggest to add a separate prop: `params` (or `resizeParams` or some better name)  of the type: `LineElementProps | PolylineAndShapeElementsProps`

src/components/Whiteboard/ElementBehaviors/Resizable/ResizeElement.tsx
```diff
@@ -70,24 +72,35 @@ export function ResizeHandleWrapper({
...
-      handlePosition={handlePosition}
       onDrag={handleDrag}
       onDragStart={onDragStart}
       onDragStop={onDragStop}
+      elementKind={rest.elementKind}
+      handlePosition={rest.handlePosition}
+      containerWidth={rest.containerWidth}
+      containerHeight={rest.containerHeight}
     />
   );
```
maheichyk, line 105: If `resizeParams` is a separate prop no need to have this `if`

src/components/Whiteboard/ElementBehaviors/Resizable/ResizeElement.tsx
```diff
@@ -191,35 +195,57 @@ export function ResizeElement({ elementId }: ResizeElementProps) {
...
 
+  if (element.kind === 'line') {
+    const renderProperties = getRenderProperties(element);
+
+    return (
+      <>
+        <ResizeHandleWrapper
+          handlePosition="start"
+          handlePositionX={renderProperties.points.start.x}
+          handlePositionY={renderProperties.points.start.y}
```
maheichyk, line 228: For separete prop can changed to:
```
resizeParams={{
  elementKind: 'line',
  handlePosition: 'start',
  handlePositionX: renderProperties.points.start.x,
  handlePositionY: renderProperties.points.start.y,
}}
```

maheichyk (changes requested):

src/components/Whiteboard/ElementBehaviors/Resizable/ResizeHandle.tsx
maheichyk: `handlePosition` could be removed, it is already defined in line 34

src/components/Whiteboard/ElementBehaviors/Resizable/types.ts
```diff
@@ -34,4 +60,15 @@ export type DimensionsHorizontal = {
   width: number;
 };
 
-export type Dimensions = DimensionsVertical & DimensionsHorizontal;
+export type DimensionsBase = DimensionsVertical & DimensionsHorizontal;
```
maheichyk, line 63: no need to export `DimensionsBase`
```suggestion
type DimensionsBase = DimensionsVertical & DimensionsHorizontal;
```

PR #361, opened 2024-03-08
Fix whiteboard initialization when whiteboards loading is slower

maheichyk:
If loading of whiteboards taking some more time, this could make whiteboard initialization to work wrong, this PR fixes the issue.

### DIFF
--- a/.changeset/shiny-beans-invite.md
+++ b/.changeset/shiny-beans-invite.md
@@ -1,0 +1,5 @@
+---
+'@nordeck/matrix-neoboard-widget': patch
+---
+
+Fix whiteboard initialization when whiteboards loading is slower.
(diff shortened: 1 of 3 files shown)


charlynguyen (approved):

.changeset/shiny-beans-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+'@nordeck/matrix-neoboard-widget': patch
+---
+
+Fix whiteboard initialization when whiteboards loading is slower.
```
weeman1337, line 5: If this was caused by another issue before doing the release, should we actually mention it,
because the issue doesn't appear in any of the releases?

weeman1337 (commented):
Looks reasonable. Though one question about the changelog item.

.changeset/shiny-beans-invite.md
maheichyk: I believe it is better to have a changelog entry in this case, even though there was no release. For transparency reasons and because it is a separate change.

weeman1337 (approved):

PR #370, opened 2024-03-19
Shape selection performance issues

mgcm:
This PR improves overall selection performance by only getting the active elements and the element overrides on the `WhiteboardHost` and passing them down to each element shape, instead of having each shape do that.

As the video shows, in this test board with 181 shapes, drag selection stutters and stops for 15 seconds when the selection range reaches the rectangle stack on the far right. Also, changing the color for all elements takes aprox. 7 seconds.

After these changes, the selection is dramatically improved and changing color for all elements is also immediate.

https://github.com/nordeck/matrix-neoboard/assets/26399/ed0f7913-3e16-4a7c-8e68-4a5179e58e18


src/components/Whiteboard/ElementBehaviors/Selection/types.ts
```diff
@@ -14,8 +14,18 @@
...
   readOnly: boolean;
   active: boolean;
   elementId: string;
 };
+
+export type WithExtendedSelectionProps = {
+  readOnly: boolean;
+  active: boolean;
+  elementId: string;
+  elementIds: string[];
```
weeman1337, line 29: I would make `elementIds` and `overrides` optional and set a default in the components.
Then we save `elementIds={[]}` in some places.

Plus `elementIds` should eventually be named `activeElementIds` instead.

src/components/Whiteboard/ElementBehaviors/Selection/types.ts
```diff
@@ -14,8 +14,18 @@
...
 
+import { Elements } from '../../../../state/types';
+
 export type WithSelectionProps = {
   readOnly: boolean;
   active: boolean;
   elementId: string;
 };
+
+export type WithExtendedSelectionProps = {
```
weeman1337, line 25: This should be an intersection type with `WithSelectionProps`.

weeman1337 (changes requested):
Some smaller comments in code.
Tests showed an issue:

Given you have multiple elements on the board
When you select one of them
And move it around
Then all elements move around

src/components/Whiteboard/ElementBehaviors/Selection/types.ts
mgcm: Doh, leveling up my TS skills, thanks!

src/components/Whiteboard/ElementBehaviors/Selection/types.ts
mgcm: Yup, makes sense

src/components/Whiteboard/ElementBehaviors/Moveable/MoveableElement.tsx
```diff
@@ -46,11 +43,15 @@ const DraggableGroup = styled('g')({
...
+  overrides?: Elements;
 }>;
 
-export function MoveableElement({ children, elementId }: MoveableElementProps) {
-  const { activeElementIds } = useActiveElements();
-  const elements = useElementOverrides(activeElementIds);
+export function MoveableElement({
+  children,
+  elementId,
+  overrides = {},
```
weeman1337, line 52: This can become

```suggestion
  overrides: elements = {},
```

then the assignment `const elements = overrides;` can go away.

mgcm: this is what I went for.

weeman1337 (approved):
Go :rocket:

PR #382, opened 2024-04-04
Support multiselect resizing

charlynguyen:

https://github.com/nordeck/matrix-neoboard/assets/1422657/634b29e0-7922-4cc5-af10-2f2490a25bdc

### DIFF
--- a/src/components/Whiteboard/ElementBehaviors/Resizable/ResizeElement.tsx
+++ b/src/components/Whiteboard/ElementBehaviors/Resizable/ResizeElement.tsx
@@ -21,7 +21,9 @@ import {
   useWhiteboardSlideInstance,
 } from '../../../../state';
 import {
-  useElementOverride,
+  ElementOverrideUpdate,
+  createResetElementOverrides,
+  useElementOverrides,
   useSetElementOverride,
 } from '../../../ElementOverridesProvider';
 import { useLayoutState } from '../../../Layout';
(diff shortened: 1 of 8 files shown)


src/components/Whiteboard/ElementBehaviors/Resizable/ResizeElement.tsx
```diff
@@ -21,336 +21,275 @@ import {
...
   );
 
-  if (!element) {
+  if (activeElements.length === 0) {
     return null;
   }
 
-  if (element.type === 'path' && element.kind === 'line') {
-    const renderProperties = getRenderProperties(element);
+  if (
```
weeman1337, line 152: Is this the case where only one line is selected?
This could use a comment to quickly understand it.

src/components/Whiteboard/ElementBehaviors/Resizable/ResizeHandle.tsx
```diff
@@ -148,26 +151,27 @@ export type DragEvent = {
 };
 
 export type ResizeHandleProps = {
+  handleProperties: HandleProperties;
```
weeman1337, line 154: At least to me it looks a bit strange that `ResizeHandleProps` has a property for the handle.
Are not all of these then handle properties?

weeman1337 (changes requested):
Tested and works :+1: 

If I got it correctly the resize is done for every element is done in a loop + the code now lives in one function inside of utils?

I've only some smaller notes about tests and probably some comments in the code for clarification, so that it is easily understandable.

I've left a comment about images in the tickets.

src/components/Whiteboard/ElementBehaviors/Resizable/ResizeHandle.tsx
charlynguyen: Good catch 🙂 I had the same thought and somehow sticked to that generic term. But it is basically all about the handle positioning (name and coords). I will rename it.

weeman1337: Looks better now, thanks :)

PR #405, opened 2024-04-22
Fix TextElement size calculation in triangle shapes

charlynguyen:

### DIFF
--- a/src/components/elements/triangle/getRenderProperties.ts
+++ b/src/components/elements/triangle/getRenderProperties.ts
@@ -35,10 +35,10 @@ export function getRenderProperties(
   const height = shape.height;
 
   // Based on https://puzzling.stackexchange.com/a/40221
-  const fitSquareLength = (width * height) / (width + height);
+  const fitSquareLength = (width * height) / (width + height) || 0;
   const paddingTop = height - fitSquareLength;
   const horizontalPadding = (width - fitSquareLength) / 2;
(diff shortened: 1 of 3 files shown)


src/components/elements/triangle/getRenderProperties.ts
```diff
@@ -35,10 +35,10 @@ export function getRenderProperties(
   const height = shape.height;
 
   // Based on https://puzzling.stackexchange.com/a/40221
-  const fitSquareLength = (width * height) / (width + height);
+  const fitSquareLength = (width * height) / (width + height) || 0;
```
weeman1337, line 38: This would end up with `Infinity` when `width + height` is zero. Does it make sense with a bottom padding of 10px? I would expect everything to be `0` here.

Can you check it and or explain?

charlynguyen: The bug is caused by us doing division by 0 (due to what you just mentioned above). The change here is to return 0 in that case.

The `10px` is some stylistic issue that causes negative values errors on the `foreignObject`. Unrelated to the bug fix. I didn't want to touch it, but somehow I feel it does not hurt to have it at least fixed for the triangle.

weeman1337: Never mind. I've read the precedence table from the wrong side :laughing:

weeman1337 (approved):

PR #434, opened 2024-05-16
Set whiteboard custom height

maheichyk:
This PR allows to set custom whiteboard height

### DIFF
--- a/packages/react-sdk/src/components/Layout/Layout.tsx
+++ b/packages/react-sdk/src/components/Layout/Layout.tsx
@@ -53,7 +53,12 @@ const TabPanelStyled = styled(TabPanel)(() => ({
   },
 }));
 
-export function Layout() {
+export type LayoutProps = {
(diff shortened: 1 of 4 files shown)


packages/react-sdk/src/components/Layout/Layout.tsx
```diff
@@ -53,7 +53,12 @@ const TabPanelStyled = styled(TabPanel)(() => ({
   },
 }));
 
-export function Layout() {
+export type LayoutProps = {
+  // Height of component, default to '100vh'
+  height?: number | string;
```
weeman1337, line 58: What does a number mean here? Should it not always include a unit?

maheichyk: Could be just a string, was not sure about type, checked some mui code and saw number that would end up with pixels, so used `number | string` there.

weeman1337 (approved):
Looks good :+1:

PR #436, opened 2024-05-16
Use early return to simplify ElementBarWrapper

weeman1337:
While working on the element bar glitch fix reading the component was hard.
It always looked like that the values could be wrong.

Using an early return simplifies the code and makes clear that nothing happens if the conditions are not met.


mgcm (approved):
improvement! 🚀

PR #486, opened 2024-08-02
Add insert slide action

weeman1337:
Adds a new context menu item „Insert slide“, that inserts a slide after the clicked slide.

insert-slide

### DIFF
--- a/packages/react-sdk/src/components/SlideOverviewBar/withContextMenu.tsx
+++ b/packages/react-sdk/src/components/SlideOverviewBar/withContextMenu.tsx
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import AddIcon from '@mui/icons-material/Add';
 import ContentCopyIcon from '@mui/icons-material/ContentCopy';
 import DeleteOutlinedIcon from '@mui/icons-material/DeleteOutlined';
 import LockOutlinedIcon from '@mui/icons-material/LockOutlined';
@@ -98,6 +99,12 @@ export const withContextMenu = <P extends object>(
       handleClose();
     }, [handleClose, slideId, whiteboardInstance]);
 
+    const handleInsertSlide = useCallback(() => {
+      const slideId = whiteboardInstance.addSlide(slideIndex - 1);
+      whiteboardInstance.setActiveSlideId(slideId);
+      handleClose();
+    }, [handleClose, slideIndex, whiteboardInstance]);
+
     const handleContextMenu = useCallback(
       (event: MouseEvent<HTMLButtonElement>) => {
         event.preventDefault();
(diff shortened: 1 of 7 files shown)


packages/react-sdk/src/components/SlideOverviewBar/withContextMenu.tsx
```diff
@@ -98,6 +99,12 @@ export const withContextMenu = <P extends object>(
       handleClose();
     }, [handleClose, slideId, whiteboardInstance]);
 
+    const handleInsertSlide = useCallback(() => {
+      const slideId = whiteboardInstance.addSlide(slideIndex - 1);
```
mgcm, line 103: This should be `slideIndex + 1` - current implementation (from video capture) seemed to work because you opened the context menu while on top of the first slide but actually it was the second slide that was selected.

I would also change `slideId` to `newSlideId` so it doesn't get confused with the component property with the same name.

packages/react-sdk/src/state/whiteboardInstanceImpl.test.ts
```diff
@@ -283,6 +283,20 @@ describe('WhiteboardInstanceImpl', () => {
...
+      mockWhiteboard(),
+      '@user-id',
+    );
+
+    const slide1 = whiteboardInstance.addSlide();
+    const slide2 = whiteboardInstance.addSlide(1);
+
+    expect(whiteboardInstance.getSlideIds()).toEqual([slide0, slide2, slide1]);
+  });
+
```
mgcm, line 299: I would also suggest we include another test:

```
it('should add new slides between existing slides', async () => {
    const whiteboardInstance = new WhiteboardInstanceImpl(
      synchronizedDocument,
      communicationChannel,
      mockWhiteboard(),
      '@user-id',
    );

    const slide1 = whiteboardInstance.duplicateSlide(slide0);
    const slide2 = whiteboardInstance.duplicateSlide(slide1);
    const newSlide1 = whiteboardInstance.addSlide(1);
    const newSlide3 = whiteboardInstance.addSlide(3);

    expect(whiteboardInstance.getSlideIds()).toEqual([slide0, newSlide1, slide1, newSlide3, slide2]);
  });
  ```

mgcm (changes requested):

packages/react-sdk/src/components/SlideOverviewBar/withContextMenu.tsx
weeman1337: > This should be slideIndex + 1 - current implementation (from video capture) seemed to work because you opened the context menu while on top of the first slide but actually it was the second slide that was selected.

I don't get it. It should add the new slide after the slide, whose context menu I have clicked?

mgcm (approved):

PR #487, opened 2024-08-02
Revert the resize handles not on top fix.

weeman1337:
The other fix changed the order of elements.
That lead to the issue, that text can't be added to a shape any more, because the selection was on top of the shape.

**:heavy_check_mark: Checklist**

- [ ] A changeset describing the change and affected packages ([more info](https://github.com/nordeck/.github/blob/main/docs/CONTRIBUTING.md#changelog-and-versioning)).
- [ ] Added or updated documentation.
- [ ] Tests for new functionality and regression tests for bug fixes.
- [ ] Screenshots or videos attached (for UI changes).
- [x] All your commits have a `Signed-off-by` line in the message ([more info](https://github.com/nordeck/.github/blob/main/docs/CONTRIBUTING.md#dco)).

PR #489, opened 2024-08-02
Render the resize handles on top

HarHarLinks:
fixes the regression while keeping the original fix instead of reverting as proposed in #487.

painting the borders individually prevents the shape's (invisible) volume blocking the needed click-space; the latter caused the regression.

  - see #485
  - see #485


weeman1337 (approved):
Tested. Fix works without breaking the ability of interacting with the elements :+1:

PR #496, opened 2024-08-07
Fix i18n error

weeman1337:
To be honest I don't know why it break.

But we have created a language detector by inheritance instead of composition like described int he i18next docs https://github.com/i18next/i18next-browser-languageDetector?tab=readme-ov-file#adding-own-detection-functionality

This PR copies the code from the widget toolkit to get it work https://github.com/nordeck/matrix-widget-toolkit/blob/main/packages/mui/src/i18n/i18n.ts#L43

There will be a follow up ticket to find a better solution (like refactor the toolkit) for it.


jaller94 (approved):
Great fix.
Let's remove create-react-scripts soon for a more permanent solution.

PR #518, opened 2024-08-15
Ensure our minimal svg size is 400px on the shortest side

MTRNord:
Fixes that svg images without size information broke. It makes the shortest side 400px while keeping the aspect ratio.

### DIFF
--- a/packages/react-sdk/src/lib/determineImageSize.ts
+++ b/packages/react-sdk/src/lib/determineImageSize.ts
@@ -31,7 +31,43 @@ export async function determineImageSize(
     const handleLoad = () => {
       image.removeEventListener('error', handleError);
       URL.revokeObjectURL(image.src);
-      resolve({ width: image.width, height: image.height });
+
+      let width = image.width;
+      let height = image.height;
+      // Ensure the if this is an SVG image the minimum shortest side is 400px
+      if (type === 'image/svg+xml' && width === 0 && height === 0) {
+        // The ratio cant be calculated if both dimensions are 0
+        // See alsl https://github.com/whatwg/html/issues/3510 on why it is important to actually render it on the DOM to move on
+        // If the image is not rendered on the DOM the image will not be loaded and the dimensions will be 0.
+
+        // Needed to calculate the aspect ratio
+        image.width = 1000;
+        // Position fixed so that the image doesn't affect layout while rendering
+        image.style.position = 'fixed';
+        // Make invisible so the image doesn't briefly appear on the screen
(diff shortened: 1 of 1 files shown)


packages/react-sdk/src/lib/determineImageSize.ts
```diff
@@ -31,7 +31,28 @@ export async function determineImageSize(
     const handleLoad = () => {
       image.removeEventListener('error', handleError);
       URL.revokeObjectURL(image.src);
-      resolve({ width: image.width, height: image.height });
+
+      let width = image.width;
+      let height = image.height;
+      // Ensure the if this is an SVG image the minimum shortest side is 400px
+      if (type === 'image/svg+xml') {
```
weeman1337, line 38: Can you merge the two ifs?

weeman1337 (changes requested):
I am unfortunately rejecting this PR :(

If you upload the example from the ticket it destroys the aspect ratio (at least when using Firefox). It should keep it. I think we can do better. Browsers are pretty good in rendering crap :D

Something like this returns a width and height for me:

```
          const measureElement = document.createElement('IMG');
          measureElement.id = 'measure-element';
          measureElement.src = URL.createObjectURL(
            new Blob([imageData], { type }),
          );
          document.body.appendChild(measureElement);
```

It still needs to be fit into the 400 max.

Tests with an SVG element specifying a width and height works. I have attached a test image to the ticket.

MTRNord:
Hm that's weird. Because an image without any size info apart from the viewport ich worked in Firefox for me while testing. I wonder if I broke something just before pushing or while cleaning things up. 

The existing image used should do the exact same as your code does but headless. 

Will investigate. 

> I am unfortunately rejecting this PR :(
> 
> If you upload the example from the ticket it destroys the aspect ratio (at least when using Firefox). It should keep it. I think we can do better. Browsers are pretty good in rendering crap :D
> 
> Something like this returns a width and height for me:
> 
> ```
>           const measureElement = document.createElement('IMG');
>           measureElement.id = 'measure-element';
>           measureElement.src = URL.createObjectURL(
>             new Blob([imageData], { type }),
>           );
>           document.body.appendChild(measureElement);
> ```
> 
> This is only example code, proving that it works. It still needs some work and fit into the 400 max.
> 
> Tests with an SVG element specifying a width and height works. I have attached a test image to the ticket.

weeman1337 (dismissed):
Tested, works :heavy_check_mark:

weeman1337 (approved):

PR #529, opened 2024-08-21
Allow setting a container for the fullscreen mode

MTRNord:
Allows setting the container of the neoboard. This is especially needed when we do not use an iframe for embedding it.

### DIFF
--- a/packages/react-sdk/src/App.tsx
+++ b/packages/react-sdk/src/App.tsx
@@ -54,5 +54,9 @@ export const App = ({ layoutProps }: AppProps) => {
     );
   }
 
-  return <Layout {...layoutProps} />;
+  return (
+    <div id="widget-root">
+      <Layout {...layoutProps} />
+    </div>
+  );
 };
--- a/packages/react-sdk/src/App.tsx
+++ b/packages/react-sdk/src/App.tsx
@@ -54,9 +54,5 @@ export const App = ({ layoutProps }: AppProps) => {
     );
   }
 
-  return (
-    <div id="widget-root">
-      <Layout {...layoutProps} />
-    </div>
-  );
+  return <Layout {...layoutProps} />;
 };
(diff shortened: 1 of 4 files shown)


packages/react-sdk/src/App.tsx
```diff
@@ -54,5 +54,9 @@ export const App = ({ layoutProps }: AppProps) => {
     );
   }
 
-  return <Layout {...layoutProps} />;
+  return (
+    <div id="widget-root">
```
weeman1337, line 58: Can we stick the ID to some element that already exist?

packages/react-sdk/src/components/Layout/useFullscreenMode.ts
```diff
@@ -39,8 +39,10 @@ export function useFullscreenMode(): UseFullScreenModeResult {
   );
 
   const setFullscreenMode = useCallback((value: boolean) => {
+    const container =
```
weeman1337, line 42: Even it is cheap, `container` is only used inside the `if`-branch. Could be moved to inside of it.

weeman1337 (approved):

packages/react-sdk/src/App.tsx
MTRNord: will need to check that. In theory yes

MTRNord: Yes this worked

packages/react-sdk/src/components/Layout/Layout.tsx
```diff
@@ -77,7 +79,11 @@ export function Layout({ height = '100vh' }: LayoutProps) {
     <SlidesProvider>
       <GuidedTour disabled={isViewingPresentation} />
 
-      <Stack height={height} direction="row" bgcolor="background.paper">
+      <Stack
+        height={!isFullscreenMode ? height : '100vh'}
```
MTRNord, line 83: This ensures we dont get a black bar under the widget if its smaller than 100vh normally.

weeman1337 (approved):
Works :heavy_check_mark: 

Seems we have another issue with integrating the browser full screen mode and the NeoBoard full screen mode :woman_shrugging:

MTRNord:
> Works ✔️
> 
> Seems we have another issue with integrating the browser full screen mode and the NeoBoard full screen mode 🤷‍♀️

You mean F11?

PR #533, opened 2024-08-26
Make F11 work for fullscreen of the widget

MTRNord:

### DIFF
--- a/packages/react-sdk/src/components/Layout/useFullscreenMode.ts
+++ b/packages/react-sdk/src/components/Layout/useFullscreenMode.ts
@@ -64,6 +64,21 @@ export function useFullscreenMode(): UseFullScreenModeResult {
     };
   }, [setFullscreenModeState]);
 
+  useEffect(() => {
+    function listener(e: KeyboardEvent) {
+      if (e.key === 'F11') {
+        e.preventDefault();
+        setFullscreenMode(!isFullscreenMode);
+      }
+    }
+
+    window.addEventListener('keydown', listener);
+
+    return () => {
+      window.removeEventListener('keydown', listener);
+    };
+  }, [setFullscreenMode, isFullscreenMode]);
+
   return {
     isFullscreenMode,
     setFullscreenMode,
--- a/packages/react-sdk/src/components/Layout/useFullscreenMode.ts
+++ b/packages/react-sdk/src/components/Layout/useFullscreenMode.ts
@@ -14,7 +14,10 @@
  * limitations under the License.
  */
 
-import { useCallback, useEffect, useState } from 'react';
+import loglevel from 'loglevel';
+import { useCallback, useEffect, useRef, useState } from 'react';
+import { useHotkeys } from 'react-hotkeys-hook';
+import { HOTKEY_SCOPE_GLOBAL } from '../WhiteboardHotkeysProvider';
 
 type UseFullScreenModeResult = {
   /**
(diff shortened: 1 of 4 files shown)


packages/react-sdk/src/components/Layout/useFullscreenMode.ts
```diff
@@ -64,6 +64,21 @@ export function useFullscreenMode(): UseFullScreenModeResult {
     };
   }, [setFullscreenModeState]);
 
+  useEffect(() => {
+    function listener(e: KeyboardEvent) {
```
weeman1337, line 68: We use [React Hotkeys Hook](https://react-hotkeys-hook.vercel.app/) in other places. It should give us more control over hotkeys in different context (such as only enable them in specific places).

Can you switch to it?

weeman1337 (changes requested):

packages/react-sdk/src/components/Layout/useFullscreenMode.ts
MTRNord: @weeman1337 I tried but it seems like either it works 2 times and then gets stuck in `TypeError: Fullscreen request denied` in firefox or changing to keyup makes it work once and then not trigger again. So something seems wonky with this hook which makes it unreliable compared to using a classic keyhandler

MTRNord: The hook seems to cause mainly `Anfrage für Vollbildmodus wurde abgelehnt, weil Element.mozRequestFullScreen() nicht aus einem kurz laufenden Benutzer-generierten Ereignis-Handler aufgerufen wurde.`

MTRNord: It seems like it works when I actively click on the board in between but it doesn't just with the keybind despite https://developer.mozilla.org/en-US/docs/Web/Security/User_activation claiming keydown is accepted. I can only assume that firefox fails due to some indirection or memo thing here

HarHarLinks (commented):
It works. I don't have anything to add about the technical discussion in the other thread in either direction. I am not sure if the implemented behavior is required or desirable.

packages/react-sdk/src/components/Layout/useFullscreenMode.ts
weeman1337: Then lets keep it like it. Adding a comment

weeman1337: Also getting this without the hotkeys hook… Seems like it is triggered twice.

weeman1337: I think this can be caused by Reacts strict mode.

weeman1337: Or the hook being used in two places.

weeman1337:
I did update the PR:
- It now has error handling
- The hook is used in one place only (and it should really only used from layout state)
- It uses react hotkeys
- It remembers the internal state, whether there is already an ongoing full-screen request 

With that changes, it works on my machine™

@HarHarLinks can you re-test it? Also in different browsers.

weeman1337 (approved):

HarHarLinks:
I can do that once there is a product decision whether this is actually desireable behavior.

HarHarLinks:
This PR changes behavior such that pressing F11 may switch it into fullscreen mode. This has the following issues:

Because Element Web/Desktop, as the currently only client really implementing widgets, hosts its widgets in iframes, it causes inconsistent behavior in browsers. Depending on the current focus, a different part of the app will capture the keyboard shortcut. If the user has most recently used the Element UI e.g. to send a message or navigate to a room, or interacted with another widget (e.g. in the NeoToolSuite), then the focus is outside the NeoBoard and pressing F11 will maximize Element or potentially the other widget instead of NeoBoard. Only when the last interaction was in NeoBoard, then this will work as a user might expect. Similar issues arise for other apps using the NeoBoard react-sdk, though the situation may be more or less severe than in the case of the widget.

Further, F11 is one of multiple de-facto standards on the web that puts the page into fullscreen, hiding the tab and address bars etc. It does not put some selected part of the page only into fullscreen. The proposal is breaking with that convention.

For comparison of purely the UX component: if I'm watching a video on YouTube, regardless where if I've last clicked on the player or the page, F11 always maximizes the page, not the player. (The custom shortcut <kbd>f</kbd> in the meantime is subject to the technical limitations at play here.)

Therefore, we reject this for now.

PR #551, opened 2024-09-05
Use Widget API for image downloads

mgcm:
Uses the recent [downloadFile](https://github.com/matrix-org/matrix-widget-api/blob/a294bb6aae8244cb7e6fcffd88e30e214830a01b/src/driver/WidgetDriver.ts#L356) Widget API with support for authenticated media.

Has been tested with:
 - the latest `develop` Element Web + React SDK, which already implements it in the Widget Driver, with authenticated media **enabled**
- the latest `develop` Element Web + React SDK, **without** authenticated media enabled on the homeserver
- previous releases of Element Web + React SDK, in which the widget falls back to using the unauthenticated Media API due to lack of support for `downloadFile` in the Widget driver

### DIFF
--- a/.changeset/fifty-crabs-attend.md
+++ b/.changeset/fifty-crabs-attend.md
@@ -1,0 +1,6 @@
+---
+'@nordeck/matrix-neoboard-widget': patch
+'@nordeck/matrix-neoboard-react-sdk': patch
+---
+
+Use the Widget API for image downloads when available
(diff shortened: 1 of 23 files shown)


HarHarLinks (dismissed):
lgtm.

i also checked the unhappy case to see that it does not break too horribly in element without widget api update but with authenticated-media-only server. it looks as usual when media is not available:
image

mgcm:
@HarHarLinks I've improved the fallback logic and this is now ready for a final review before releasing 1.19.0

MTRNord (approved):
Only tested function. Seems to work with up and downloads in an encrypted room on a authed media only server

HarHarLinks (approved):
awesome!

PR #581, opened 2024-09-24
Deactivate grid in presentation mode

ahmadkadri:
Deactivate grid in presentation mode and return the same value after ending the presentation.

https://github.com/user-attachments/assets/843ae0ee-dd9e-4a45-b231-b380a4c5b0cc


HarHarLinks (changes requested):
needs a changeset

HarHarLinks (approved):
Strictly speaking did not meet the ACs, but after clarification is accepted.

weeman1337 (commented):
Please wait, need to do a test.

weeman1337 (changes requested):
Doesn't work for me. I will do some tests…

packages/react-sdk/src/components/PresentBar/PresentBar.tsx
```diff
@@ -56,6 +58,23 @@ export function PresentBar() {
   const isPresenting = state.type === 'presenting';
   const isPresentingInEditMode = isPresenting && state.isEditMode;
 
+  const togglePresentationMode = useCallback(() => {
+    const storedGridStatus = localStorage.getItem('showGridState');
+    if (!isPresenting) {
+      if (isShowGrid) {
+        localStorage.setItem('showGridState', 'true');
```
weeman1337, line 65: We should also unset the value here, if the grid is off. When you leave the board without stopping the presentation with the button, then the old value is still in local-storage:

- turn grid on
- press presentation button (start presentation)
- leave the room
- go back to room
- turn grid off
- press presentation button (start presentation)
- press presentation button (end presentation)

Expected: No grid

What happens: grid

packages/react-sdk/src/components/PresentBar/PresentBar.tsx
```diff
@@ -56,6 +58,23 @@ export function PresentBar() {
...
+        setShowGrid(false);
+      }
+      togglePresentation();
+    } else {
+      if (storedGridStatus === 'true') {
+        setShowGrid(true);
+        localStorage.removeItem('showGridState');
+      }
+      togglePresentation();
```
weeman1337, line 74: It call `togglePresentation()` in both cases. It could be moved outside of the if statements at the end of the callback. Then also the if conditions can be combined, so that we get less if-blocks.

packages/react-sdk/src/components/PresentBar/PresentBar.tsx
```diff
@@ -56,6 +58,17 @@ export function PresentBar() {
   const isPresenting = state.type === 'presenting';
   const isPresentingInEditMode = isPresenting && state.isEditMode;
 
+  const togglePresentationMode = useCallback(() => {
+    const storedGridStatus = localStorage.getItem('showGridState');
+    if (!isPresenting) {
+      localStorage.setItem('showGridState', isShowGrid ? 'true' : 'fasle');
```
weeman1337, line 64: `fasle`? :D → `false`

weeman1337 (changes requested):
Small typo, other than that it is good to go :rocket:

weeman1337 (approved):
LGTM

PR #588, opened 2024-09-30
Hide about menu item if the widget is embedded

ahmadkadri:


packages/react-sdk/src/components/HelpCenterBar/HelpMenu.tsx
```diff
@@ -29,6 +29,7 @@ export function HelpMenu() {
   const helpCenterUrl = getEnvironment('REACT_APP_HELP_CENTER_URL');
   const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
   const { restartGuidedTour } = useGuidedTour();
+  const embedded = getEnvironment('REACT_APP_EMBADDED') === 'true';
```
weeman1337, line 32: Small typo

```suggestion
  const embedded = getEnvironment('REACT_APP_EMBEDDED') === 'true';
```

weeman1337 (changes requested):
:thinking: I am not sure if an environment variable is the way to go. But I am happy to discuss it.

We already pass props to the React SDK app: https://github.com/nordeck/matrix-neoboard/blob/main/packages/react-sdk/src/App.tsx#L23 I would have expected a new prop there, that can then be used to tell the NeoBoard application whether it is embedded. Eventually this can be just another „layout prop” since hiding something is also „layout“? :D

mgcm:
> 🤔 I am not sure if an environment variable is the way to go. But I am happy to discuss it.

When I discussed this PR with Ahmad, this was an option we considered as well.

The current status is that we have a couple more env vars that configure the widget, such as `REACT_APP_HELP_CENTER_URL` and `REACT_APP_HOME_SERVER_URL`, so given the low effort of adding yet another env var, it seemed reasonable.

I think having an `embedded` property in LayoutProps is a good option but as we are still discovering what new behaviours we should adjust for in the Widget, maybe this can slide thru as is?

I mean, will `embedded` only affect the canvas layout? Maybe it will have other non-layout related changes associated with it in the future..

Regardless of how it's implemented, what is missing in this PR is the documentation of how this flag affects the widget.

weeman1337 (changes requested):
Please add a documentation about the env var here https://github.com/nordeck/matrix-neoboard/blob/main/docs/configuration.md

docs/configuration.md
```diff
@@ -17,6 +17,9 @@ REACT_APP_HELP_CENTER_URL="https://github.com/nordeck/matrix-neoboard"
 # optional - enable connection to React standalone devtools
 # Does only work in development mode.
 REACT_APP_DEVTOOLS=true
+
+# Indicate if the widget is embedded in standalone mode
```
weeman1337, line 21: We should document what it means. In particular hide the about button.

weeman1337 (changes requested):

weeman1337 (approved):
Looks good now :+1:

PR #591, opened 2024-10-07
Expose whiteboard destroy and persist APIs

weeman1337:
Required for save-on-leave feature


packages/react-sdk/src/state/synchronizedDocumentImpl.test.ts
```diff
@@ -707,6 +707,94 @@ describe('SynchronizedDocumentImpl', () => {
...
+          snapshotOutstanding: true,
+          snapshotsReceived: 0,
+          snapshotsSend: 0,
+        },
+      ]);
+
+      await isReady;
+
+      await waitFor(() => {
+        expect(widgetApi.sendRoomEvent).toHaveBeenCalledTimes(2);
```
mgcm, line 763: Feels awkward to check for generic sending of any type of room event when the test is about snapshot events.

I realize that other previous tests in this suite had this same approach but I also see that we can actually check for event types if we use `toHaveBeenCalledWith`.

But since we care about the number of calls, maybe complement it with `toHaveBeenLastCalledWith`?

Anyway, not a blocker, just thinking if / how it can be improved.

packages/react-sdk/src/state/synchronizedDocumentImpl.test.ts
```diff
@@ -707,6 +707,94 @@ describe('SynchronizedDocumentImpl', () => {
...
+
+      await firstValueFrom(
+        synchronizedDocument
+          .observeIsLoading()
+          .pipe(filter((loading) => !loading)),
+      );
+
+      synchronizedDocument.persist();
+
+      expect(widgetApi.sendRoomEvent).not.toHaveBeenCalled();
```
mgcm, line 795: Ah but here is a good candidate for `toHaveBeenCalledWith`.

packages/react-sdk/src/state/synchronizedDocumentImpl.ts
```diff
@@ -149,6 +149,11 @@ export class SynchronizedDocumentImpl<T extends Record<string, unknown>>
         // deduplicate the requests in case the snapshot creation is slower than 5000ms
         throttle(
           async ({ doc }) => {
+            if (!this.statistics.snapshotOutstanding) {
```
mgcm, line 152: Would it make sense to DRY this and just call the new `persist` method here? It's essentially the same code except for the try / catch.

packages/react-sdk/src/state/synchronizedDocumentImpl.test.ts
weeman1337: It should **not** have been called at all. So no specific test necessary here.

packages/react-sdk/src/state/synchronizedDocumentImpl.ts
weeman1337: Good point, refactored :hammer_and_pick:

packages/react-sdk/src/state/synchronizedDocumentImpl.test.ts
weeman1337: Updated, it now checks for snapshot events

mgcm (approved):

PR #620, opened 2024-11-04
Implement connection monitoring

weeman1337:
User perspective:
- Monitors the network state
- Displays a snackbar on send snapshot errors
- If the user clicks „learn more“ it shows a dialog with a message depending on the network state
- Tries to resend snapshots until success

error-handling

What changed:
- Snackbar can now display a „priority“ snackbar, overriding every other
- Snackbar is no longer dismissable by clicking anywhere
- New store slice with a connection info, currently providing if a snapshot failed
- Synchronised document now updates the last snapshot state in the store
- `ConnectionStateProvider` handling the connection logic
- A separate `ConnectionStateDialog` to display the „learn more“ info
- New `ObservableBehaviorSubject` as a helper
- `WhiteboardManager` now offers a subject to get changed whiteboard updates

Limitations:
- This PR does not handle the problem, that the hosting client queues the events → follow up task
- Since there is no specific error for rate limiting, the dialog only shows one of the „no internet connection“ or „common“ messages

### DIFF
--- a/packages/react-sdk/src/components/connection-state/ConnectionStateProvider.tsx
+++ b/packages/react-sdk/src/components/connection-state/ConnectionStateProvider.tsx
@@ -16,16 +16,31 @@
 
 import WarningAmberIcon from '@mui/icons-material/WarningAmber';
 import { Box, Button, useTheme } from '@mui/material';
+import { getLogger } from 'loglevel';
 import React, {
   createContext,
   PropsWithChildren,
   useCallback,
   useEffect,
+  useMemo,
+  useRef,
   useState,
 } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useNetworkState } from 'react-use';
 import { useSnackbar } from '../../components/Snackbar';
+import { useWhiteboardManager } from '../../state';
+import { useDistinctObserveBehaviorSubject } from '../../state/useDistinctObserveBehaviorSubject';
+import { useAppDispatch } from '../../store';
+import {
+  selectConnectionInfo,
+  setSnapshotSuccessful,
+} from '../../store/connectionInfoSlice';
+import { useAppSelector } from '../../store/reduxToolkitHooks';
+
+const logger = getLogger('ConnectionStateProvider');
+
+const RETRY_INTERVAL = 10_000;
 
 type ConnectionState = 'online' | 'no_internet_connection';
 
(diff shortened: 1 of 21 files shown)


packages/react-sdk/src/components/connection-state/ConnectionStateProvider.tsx
```diff
@@ -0,0 +1,138 @@
...
+  ConnectionStateState | undefined
+>(undefined);
+
+export const ConnectionStateProvider: React.FC<PropsWithChildren> = function ({
+  children,
+}) {
+  const theme = useTheme();
+  const { t } = useTranslation();
+  const { clearSnackbar, showSnackbar, snackbarProps } = useSnackbar();
+  const networkState = useNetworkState();
```
MTRNord, line 48: It might be worth adding a comment that in I believe all browsers this only tells you if the router is unreachable. However, a working LAN without Internet/Server connection would still be silently failing here. https://developer.mozilla.org/en-US/docs/Web/API/Navigator/onLine mentions the behaviour which https://github.com/streamich/react-use/blob/e1d0cd9f7fb2a124a9d46879489abfefdf48d836/src/useNetworkState.ts#L102-L130 uses via the event listeners.

Function wise this might be ok but we should be aware of this behaviour when we look at this in $x months again.

weeman1337: Jep, you are right. The browser network connection API is not great. I have also mentioned it in the ticket already. Next PR will be active monitoring by sending requests to the homeserver.

Comment added.

packages/react-sdk/src/components/connection-state/ConnectionStateProvider.tsx
```diff
@@ -0,0 +1,142 @@
...
+>(undefined);
+
+export const ConnectionStateProvider: React.FC<PropsWithChildren> = function ({
+  children,
+}) {
+  const theme = useTheme();
+  const { t } = useTranslation();
+  const { clearSnackbar, showSnackbar, snackbarProps } = useSnackbar();
+  /**
+   * Basic connection monitoring. Works if there is any network connection and is highly browser dependant.
```
jaller94, line 49:
```suggestion
   * Basic connection monitoring. Works if there is any network connection and is highly browser-dependent.
```

jaller94 (dismissed):
Looks ok to me.

On Matrix we discussed:
* Connection issues may also affect loading but only saving is mentioned in this PR.
* The dialog also shows up when there are no changes to save.

weeman1337:
> Looks ok to me.
> 
> On Matrix we discussed:
> 
>     * Connection issues may also affect loading but only saving is mentioned in this PR.
> 
>     * The dialog also shows up when there are no changes to save.

Both should be done now

jaller94 (approved):
Looks good and you also found a fix for this working in standalone. :clap: 
https://github.com/nordeck/matrix-neoboard-standalone/pull/251

PR #626, opened 2024-11-06
Add rate limiting docs

weeman1337:

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,8 @@ The minimal Element version to use this widget is `1.11.8`.
 
 After checkout, run `yarn install` to download the required dependencies
 
-> **Warning** Do not use `npm install` when working with this package.
+> [!WARNING]
+> Do not use `npm install` when working with this package.
 
 ### Configuration
 
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 ![Build](https://github.com/nordeck/matrix-neoboard/workflows/CI/badge.svg)
 [![Matrix](https://img.shields.io/matrix/nordeck%3Amatrix.org)](https://matrix.to/#/#nordeck:matrix.org)
 
-A collaborative whiteboard widget for Element, based on Matrix.
+A collaborative whiteboard widget for Element, based on [Matrix](https://matrix.org).
 Use it to collaborate in real-time with your peers, by keeping full control of your data and keep it secure with end-to-end encryption.
 
 <table>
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -141,10 +141,18 @@ docker run --rm -e REACT_APP_HOME_SERVER_URL=https://matrix-client.matrix.org -p
...
-> Important:
-> The collaboration features require you to configure TURN servers in your Homeserver.
-> It can work without them, but there is a high chance that it won't!
-> The official matrix.org homeserver will work; see for example how to configure them for [Synapse](https://matrix-org.github.io/synapse/latest/turn-howto.html).
+### Important notes
+
+**TURN server**
+
+The collaboration features require you to configure TURN servers in your Homeserver.
+It can work without them, but there is a high chance that it won't!
```
HarHarLinks, line 149: I realise you didn't change this, but I don't like this wording. We should explain that while it might work to a limited extent, there might be issues.

(We should also make the widget good enough to be able to remove this or narrow it down to "it won't be instant, but it will still work except for $uncritical_features" ;) )

README.md
```diff
@@ -141,10 +141,18 @@ docker run --rm -e REACT_APP_HOME_SERVER_URL=https://matrix-client.matrix.org -p
 
 We also provide a [HELM chart](./charts/).
 
-> Important:
-> The collaboration features require you to configure TURN servers in your Homeserver.
-> It can work without them, but there is a high chance that it won't!
-> The official matrix.org homeserver will work; see for example how to configure them for [Synapse](https://matrix-org.github.io/synapse/latest/turn-howto.html).
+### Important notes
```
HarHarLinks, line 144: I feel like this info should not only be under deployment, because it is possible to just use the widget (e.g. the demo instance further up) and therefore never read this. Users should still be aware of these issues in these cases, don't you agree? 🤔

docs/configuration.md
```diff
@@ -31,6 +35,29 @@ User needs to have permissions to send these events to initialize a whiteboard:
...
 User will need to wait for the moderator to join and initialize the room and whiteboard if the user doesn't have these permissions.
 
-### Customization
+## Rate limiting settings
 
-More environment variables for UI customization [@matrix-widget-toolkit/mui](https://www.npmjs.com/package/@matrix-widget-toolkit/mui#customization).
+For a good NeoBoard experience rate limiting settings need to be tweaked.
+
+NeoBoard sends one snapshot every 5 seconds, if the board changed. One snapshot may consist of multiple chunk events.
+If you upload images, there is one media upload per image. Uploads also happen when importing a NeoBoard.
```
HarHarLinks, line 43:
```suggestion
If you added images to the board, there is one media upload per image. Uploads also happen when importing a NeoBoard.
```

docs/configuration.md
```diff
@@ -31,6 +35,29 @@ User needs to have permissions to send these events to initialize a whiteboard:
...
+```
+
+The calculation base for these settings is:
+
+- 1 snapshot with 10 chunks = 11 events in 5 seconds + additional messages sent by the user
+- It is possible to import a NeoBoard with 100 images
+
+You can find more information about where to set the rate limiting settings [in the Synapse configuration manual](https://element-hq.github.io/synapse/latest/usage/configuration/config_documentation.html#ratelimiting).
+
+Loosening the rate limiting settings also means people are allowed to send more messages to a room or upload more media files within a short time.
```
HarHarLinks, line 63:
```suggestion
Loosening the rate limiting settings also means people are allowed to send more messages to a room or upload more media files within a short time outside of NeoBoard, so these settings should be adjusted with care.
```

HarHarLinks (commented):
Good overall, just some suggestions

README.md
weeman1337: Same applies to turn servers then. Should we just make the import notes a top level headline?

README.md
weeman1337: Can you create another ticket for it? This PR is about rate limiting docs.

HarHarLinks:
What's wrong with the DCO?

HarHarLinks (approved):

PR #631, opened 2024-11-12
Send snapshot on join/invite

mgcm:


packages/react-sdk/src/components/FallbackSnapshotProvider/FallbackSnapshotProvider.tsx
```diff
@@ -0,0 +1,154 @@
...
+} from '../../store/api';
+
+export function FallbackSnapshotProvider({ children }: PropsWithChildren<{}>) {
+  const whiteboardInstance = useActiveWhiteboardInstance();
+  const { data: roomMembers } = useGetRoomMembersQuery();
+  const { data: roomEncryption } = useGetRoomEncryptionQuery();
+  const { data: roomHistoryVisibility } = useGetRoomHistoryVisibilityQuery();
+  const ownUserId = useWidgetApi().widgetParameters.userId;
+
+  useEffect(() => {
```
weeman1337, line 38: On load, do we already know the entire state? Means do we know the snapshots and invites? If not, we should delay the logic to give NeoBoard some time to retrieve all the data.

packages/react-sdk/src/components/FallbackSnapshotProvider/FallbackSnapshotProvider.tsx
```diff
@@ -0,0 +1,154 @@
...
+}
+
+function shouldSendFallbackSnapshot(
+  isRoomEncrypted: boolean,
+  isRoomHistoryVisibilityShared: boolean,
+  roomHistoryVisibilityValue: string | undefined,
+): boolean {
+  return (
+    isRoomEncrypted ||
+    (!isRoomHistoryVisibilityShared && roomHistoryVisibilityValue !== undefined)
```
weeman1337, line 117: Why is the value checked here? Inside `checkIfRoomHistoryVisibilityShared` it is already tested to be `shared`. I would move everything into `checkIfRoomHistoryVisibilityShared`.

packages/react-sdk/src/components/FallbackSnapshotProvider/FallbackSnapshotProvider.tsx
```diff
@@ -0,0 +1,154 @@
...
+      roomHistoryVisibilityValue,
+    );
+
+    if (sendFallbackSnapshot) {
+      const membershipFilter = getMembershipFilter(roomHistoryVisibilityValue);
+      const members = roomMembers?.entities;
+
+      if (members) {
+        const invitesOrJoins = Object.values(members).filter(membershipFilter);
+        const sortedInvitesOrJoins = invitesOrJoins.sort((a, b) => {
```
weeman1337, line 58: Can you move this into a separate function, e.g. `extractLastInviteOrJoin`?

(Also I would use `.reduce` instead, but this is optional).

packages/react-sdk/src/components/FallbackSnapshotProvider/FallbackSnapshotProvider.tsx
```diff
@@ -0,0 +1,154 @@
...
+          return b.origin_server_ts - a.origin_server_ts;
+        });
+        const lastInviteOrJoin = sortedInvitesOrJoins[0];
+
+        if (
+          shouldUseInviteTimestamp(lastInviteOrJoin, roomHistoryVisibilityValue)
```
weeman1337, line 64: Why doesn't it explode here? Technically `lastInviteOrJoin` can be `undefined`.

weeman1337 (changes requested):

packages/react-sdk/src/components/FallbackSnapshotProvider/FallbackSnapshotProvider.tsx
mgcm: well, in practice, there's always at least one room membership of type join, or else the room would be empty, no?

packages/react-sdk/src/components/FallbackSnapshotProvider/FallbackSnapshotProvider.tsx
```diff
@@ -0,0 +1,154 @@
...
+  isRoomHistoryVisibilityShared: boolean,
+  roomHistoryVisibilityValue: string | undefined,
+): boolean {
+  return (
+    isRoomEncrypted ||
+    (!isRoomHistoryVisibilityShared && roomHistoryVisibilityValue !== undefined)
+  );
+}
+
+function getMembershipFilter(
```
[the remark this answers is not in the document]
mgcm: Could be, I guess. Would this be by adding a new query to the `roomMemberApi` that filters membership events based on type and only return joins+invites? I can see how that would be optimal for rooms with high membership volume but it's also 10x more code... Maybe there's a simpler way I'm missing?

weeman1337 (changes requested):
This needs an in-depth review. Requesting change as a blocker notice.

packages/react-sdk/src/App.tsx
```diff
@@ -15,20 +15,105 @@
...
 };
 
+const logger = getLogger('App');
+
 export const App = ({ layoutProps }: AppProps) => {
   const { t } = useTranslation('neoboard');
   const { value, loading } = useOwnedWhiteboard();
   const whiteboardManager = useWhiteboardManager();
-  const ownUserId = useWidgetApi().widgetParameters.userId;
+  const whiteboardInstance = useActiveWhiteboardInstance();
```
weeman1337, line 45: This breaks NeoBoard standalone. Can you have a look?

packages/react-sdk/src/App.tsx
```diff
@@ -15,20 +15,105 @@
...
+  const widgetApi = useWidgetApi();
+  const ownUserId = widgetApi.widgetParameters.userId;
+  const { limitedHistoryVisibility, delayedPersist, lastMembershipEventTs } =
+    usePersistOnJoinOrInvite();
+  const { data: latestSnapshot } = useGetDocumentSnapshotQuery({
+    documentId: whiteboardInstance.getDocumentId(),
+    validator: isValidWhiteboardDocumentSnapshot,
+  });
+
+  const handlePersist = useCallback(() => {
```
weeman1337, line 55: Would be nice if the entire logic is split out, e.g. into a hook or function. Otherwise `App` becomes really cluttered.

PR #640, opened 2024-11-20
Preprocess svg images before the pdf export to ensure that they are displayed as expected

MTRNord:
There were some lying going on in nwb files and some svg files do not set sizes when canvas needs those.

### DIFF
--- a/packages/react-sdk/src/components/BoardBar/pdf/utils.ts
+++ b/packages/react-sdk/src/components/BoardBar/pdf/utils.ts
@@ -40,7 +40,73 @@ export function image(element: ImageElement, base64content: string): Content {
   };
 }
 
+/**
+ * This ensures that the svg is meeting the following conditions:
+ *
+ * - The width and height are set (required for the canvas to work)
+ * - The svg is actually an svg
+ *   - If the svg is not an svg we return the original string
+ *   - If the svg is detected to be a png or jpeg we return the original string but set the mimetype accordingly
+ *
+ * @param element The element that is being processed
+ * @param svg The base64 encoded svg string
+ * @returns The base64 encoded svg string after the preprocessing
+ */
+function preprocessSvg(element: ImageElement, svg: string): string {
+  console.log(element);
+  if (element.mimeType !== 'image/svg+xml') {
+    return svg;
+  }
+
+  // Convert the svg base64 to the actual svg
+  const svgString = atob(svg);
+
+  // We lied in the past about the mimetype, so we need to check if the svg is actually an svg and otherwise return the original string
+  if (!svgString.startsWith('<svg')) {
+    element.mimeType = svgString.includes('PNG') ? 'image/png' : 'image/jpeg';
+    return svg;
+  }
+
+  console.log(svgString);
+
+  // Parse the svg string
+  const parser = new DOMParser();
+  const doc = parser.parseFromString(svgString, 'image/svg+xml');
+  const svgElement = doc.documentElement;
+  console.log(svgElement);
+
+  // Check if a size is set and if not set it to the size from the viewbox
+  if (!svgElement.getAttribute('width') || !svgElement.getAttribute('height')) {
+    const viewBox = svgElement.getAttribute('viewBox')?.split(' ');
+    if (viewBox) {
+      svgElement.setAttribute('width', viewBox[2]);
+      svgElement.setAttribute('height', viewBox[3]);
+    } else {
+      svgElement.setAttribute('width', element.width.toString());
+      svgElement.setAttribute('height', element.height.toString());
+    }
+  }
+  console.log(svgElement);
+
+  // Convert the svg element to a string
+  const serializer = new XMLSerializer();
+  const svgStringProcessed = serializer.serializeToString(svgElement);
+
+  // Convert the svg string to base64
+  return btoa(svgStringProcessed);
+}
+
 export function conv2png(element: ImageElement, base64content: string): string {
+  // If we have a svg image we need to preprocess it for the canvas to work
+  if (element.mimeType === 'image/svg+xml') {
+    base64content = preprocessSvg(element, base64content);
+  }
+
+  // If we now end up with a png or jpeg image we can return it early
+  if (element.mimeType === 'image/png' || element.mimeType === 'image/jpeg') {
+    return base64content;
+  }
+
   const img = new Image();
   img.src = 'data:' + element.mimeType + ';base64,' + base64content;
 


packages/react-sdk/src/components/BoardBar/pdf/utils.ts
```diff
@@ -40,7 +40,68 @@ export function image(element: ImageElement, base64content: string): Content {
...
+}
+
 export function conv2png(element: ImageElement, base64content: string): string {
+  // If we have a svg image we need to preprocess it for the canvas to work
+  if (element.mimeType === 'image/svg+xml') {
+    base64content = preprocessSvg(element, base64content);
+  }
+
+  // If we now end up with a png or jpeg image we can return it early
+  if (element.mimeType === 'image/png' || element.mimeType === 'image/jpeg') {
```
weeman1337, line 101: The function name is `conv2png` »png«, why do we also return jpegs? Or doesn't it matter?

packages/react-sdk/src/components/BoardBar/pdf/utils.ts
```diff
@@ -40,7 +40,68 @@ export function image(element: ImageElement, base64content: string): Content {
...
+  const serializer = new XMLSerializer();
+  const svgStringProcessed = serializer.serializeToString(svgElement);
+
+  // Convert the svg string to base64
+  return btoa(svgStringProcessed);
+}
+
 export function conv2png(element: ImageElement, base64content: string): string {
+  // If we have a svg image we need to preprocess it for the canvas to work
+  if (element.mimeType === 'image/svg+xml') {
```
weeman1337, line 96: The mime type check is now doubled. Because `preprocessSvg` also explains it in it's description, maybe drop it here? No strong opinion…

packages/react-sdk/src/components/BoardBar/pdf/utils.ts
```diff
@@ -40,7 +40,68 @@ export function image(element: ImageElement, base64content: string): Content {
...
+  if (element.mimeType !== 'image/svg+xml') {
+    return svg;
+  }
+
+  // Convert the svg base64 to the actual svg
+  const svgString = atob(svg);
+
+  // We lied in the past about the mimetype, so we need to check if the svg is actually an svg and otherwise return the original string
+  if (!svgString.includes('<svg')) {
+    element.mimeType = svgString.includes('PNG') ? 'image/png' : 'image/jpeg';
```
weeman1337, line 65: Are we sure it MUST be a JPEG, if it is not a PNG? We do also allow GIF uploads (and maybe more in the future, wuch as webp etc.).

packages/react-sdk/src/components/BoardBar/pdf/utils.ts
```diff
@@ -40,7 +40,68 @@ export function image(element: ImageElement, base64content: string): Content {
...
+ * - The width and height are set (required for the canvas to work)
+ * - The svg is actually an svg
+ *   - If the svg is not an svg we return the original string
+ *   - If the svg is detected to be a png or jpeg we return the original string but set the mimetype accordingly
+ *
+ * @param element The element that is being processed
+ * @param svg The base64 encoded svg string
+ * @returns The base64 encoded svg string after the preprocessing
+ */
+function preprocessSvg(element: ImageElement, svg: string): string {
```
weeman1337, line 55: Bonus points :100: for some tests ;)

weeman1337 (changes requested):
This fixes the bug explained in the ticket.

Still some comments and questions.

packages/react-sdk/src/components/BoardBar/pdf/utils.ts
MTRNord: oops no thats a bug I wrote then :)

packages/react-sdk/src/components/BoardBar/pdf/utils.ts
MTRNord: Function is gone. Resolving

packages/react-sdk/src/components/BoardBar/pdf/utils.ts
MTRNord: Function is gone. Resolving

packages/react-sdk/src/components/BoardBar/pdf/utils.ts
MTRNord: Function is gone. Resolving

packages/react-sdk/src/components/BoardBar/pdf/utils.ts
```diff
@@ -31,7 +31,15 @@ export function canvas(element: CanvasElement): Content {
 }
 
 export function image(element: ImageElement, base64content: string): Content {
-  const dataURL = 'data:' + element.mimeType + ';base64,' + base64content;
+  // Cheap way to detect image type
+  const decoded = atob(base64content).trim();
+  let mimeType = 'image/png';
+  // Note: JFIF is the magic string in JPEG files.
+  if (decoded.includes('JFIF')) {
```
weeman1337, line 38: This would scan a huge PNG or other files. `JFIF` should also be somewhere at be beginning, right? Can we improve here?

packages/react-sdk/src/components/elements/image/ImageDisplay.tsx
```diff
@@ -46,14 +46,22 @@ const Image = styled('image', {
   visibility: loading ? 'hidden' : 'visible',
 }));
 
+function isSVG(svg: string): boolean {
+  const parser = new DOMParser();
+  const svgDom = parser.parseFromString(svg, 'image/svg+xml');
```
weeman1337, line 51: What happens, if we try to parse a PNG here?

Should this be more efficient, for example by testing if the first N bytes container `<svg` or something like that?

Also similar is done in other places. Please extract into a utility module.

packages/react-sdk/src/components/BoardBar/pdf/utils.ts
MTRNord: I am actually not entirely sure how to do this better. There is stuff (which is not deterministic) before this :(

For svg's we can just domparse things but for png and jpg I am not sure. We can use signatures on it with base64 but those fall apart with zip or tar alike formats. Not sure if we ever will encounter this but these seemed more future proof than relying on magic signatures.

weeman1337: I think we're talking past each other :D

It looks like, that `JFIF` is somewhere at the start of the file.

```sh
❯ xxd -l 100 Katzen\ Witze.jpg 
00000000: ffd8 ffe0 0010 4a46 4946 0001 0101 0048  ......JFIF.....H
00000010: 0048 0000 ffdb 0043 0006 0405 0605 0406  .H.....C........
```

`includes()` parses the file until the end to find it, I think. For large files this is quite expensive.

packages/react-sdk/src/components/BoardBar/pdf/createWhiteboardPdfElementImage.ts
```diff
@@ -19,19 +19,18 @@ import { ImageElement } from '../../../state';
...
-): Content {
+): Promise<Content> {
   const file = files.find((f) => f.mxc === element.mxc);
   if (file) {
-    // pdfmake only supports png and jpeg images,
-    // so we use a canvas to convert other image types to png
-    // ref: https://pdfmake.github.io/docs/0.1/document-definition-object/images/
-    if (element.mimeType === 'image/png' || element.mimeType === 'image/jpeg') {
+    const decoded = atob(file.data).trim();
+    if (decoded.includes('PNG') || decoded.includes('JFIF')) {
```
weeman1337, line 29: See other comment about this. We should consider the position of the magic strings. Otherwise we may have false-positives.

weeman1337, line 29: What about GIFs by the way?

packages/react-sdk/src/components/BoardBar/pdf/utils.ts
weeman1337, line 38: Why don't we need to detect PNG or GIFs here?

packages/react-sdk/src/state/crdt/documents/elements.ts
```diff
@@ -114,7 +114,10 @@ export type ImageElement = ElementBase & {
    */
   mxc: string;
   fileName: string;
-  mimeType: ImageMimeType;
+  /**
+   * @deprecated  This is kept for backwards compatibility. We dont send it anymore. DO NOT USE!
```
weeman1337, line 118: Maybe also mention it [here](https://github.com/nordeck/matrix-neoboard/blob/fd631c6b9b92bf7d04444e1a4e72c2ccd99e767f/docs/model/crdt-documents.md?plain=1#L195).

weeman1337 (changes requested):
A presentation with one image of each kind: svg, png, jpeg and gif is not exported to PFD. It fails with

```
ML Parsing Error: not well-formed
Location: https://localhost:5173/
Line Number 1, Column 11: [localhost:5173:1:11](https://localhost:5173/)
XML Parsing Error: not well-formed
Location: https://localhost:5173/
Line Number 1, Column 11: [localhost:5173:1:11](https://localhost:5173/)
uncaught exception: Invalid image: Error: Unknown image format.
Images dictionary should contain dataURL entries (or local file paths in node.js)
```

Tested on Firefox 132.0.2

weeman1337:
> We also got the presentation mode bug and it seems like its due to the blobs being already invalidated

What does it mean? That this PR will also solve that bug?

packages/react-sdk/src/components/elements/image/ImageDisplay.tsx
```diff
@@ -104,14 +104,30 @@ function ImageDisplay({
...
+    ): Promise<Blob> => {
       if (!(result.file instanceof Blob)) {
         throw new Error('Got non Blob file response');
       }
-      return result.file.slice(0, result.file.size, mimeType);
+      // Check if the blob is an SVG
+      // The try catch is because of the blob to text conversion
+      try {
+        // This can also throw. Therefor there is the outer catch
+        const stringFromBlob = await result.file.text();
```
weeman1337, line 117: Then put it into the inner `try` block? The `catch` block seems to be the same.

MTRNord:
> > We also got the presentation mode bug and it seems like its due to the blobs being already invalidated
> 
> What does it mean? That this PR will also solve that bug?

Oops this Was meant to go on my other PR 🤦 I seem to have mixed up the Tabs

weeman1337 (changes requested):
Export works, image display on the board works

Unfortunately importing the test board from the ticket fails with „unsupported filetype“.

Also a changeset is missing

weeman1337 (changes requested):
Unfortunately, now the test board from the ticket errors on import.

weeman1337 (approved):
Tested, works - go :rocket:

PR #661, opened 2024-12-05
Improve devtools more

HarHarLinks:

image


packages/react-sdk/src/components/DeveloperTools/DeveloperToolsDialog.tsx
```diff
@@ -133,11 +133,19 @@ export function DeveloperToolsDialog({
             </AccordionSummary>
             <AccordionDetails>
               <WhiteboardSessionsTable
-                sessions={communicationChannel.sessions?.filter(
-                  (s) =>
-                    s.sessionId !== communicationChannel.localSessionId &&
-                    s.expiresTs - Date.now() > 0,
-                )}
+                sessions={communicationChannel.sessions
```
weeman1337, line 136: Maybe this should now go into a `useMemo`.

packages/react-sdk/src/components/DeveloperTools/WhiteboardSessions.tsx
```diff
@@ -26,9 +27,18 @@ import {
 export function WhiteboardSessionsTable({
   sessions,
 }: {
-  sessions: SessionState[] | undefined;
+  sessions: (SessionState & { status?: string })[] | undefined;
 }) {
   const { t } = useTranslation('neoboard');
+  const [now, setNow] = useState(Date.now());
+  useEffect(() => {
```
weeman1337, line 34: Can be replaced by `useInterval` from [react-use](https://github.com/streamich/react-use/blob/master/docs/useInterval.md)

weeman1337 (changes requested):
Tested, works. Some smaller comments.

If touching this, can we fix the size:

image

The size is getting lower if you add more Elements to the board :D
https://github.com/nordeck/matrix-neoboard/pull/661/files#diff-84889a984e5272eb7cf18db4e6a51643a897fcdc19c9de40d85a0082fa12493bL72

HarHarLinks:
> can we fix the size

it is not a size. it is the ratio of binary size to json size (whatever that means exactly). i.e. if you subtract 1 it is the overhead of the whole encoding crdt etc stuff

weeman1337:
> > can we fix the size
> 
> it is not a size. it is the ratio of binary size to json size (whatever that means exactly). i.e. if you subtract 1 it is the overhead of the whole encoding crdt etc stuff
> 
> > The size is getting lower if you add more Elements to the board :D
> 
> The _overhead_ is getting lower, as expected.

„whatever that means exactly“ :shrug:

HarHarLinks:
suggest what to do with it? "fix" is not very precise

packages/react-sdk/src/components/DeveloperTools/WhiteboardSessions.tsx
HarHarLinks: done

weeman1337 (changes requested):
There is a new error message in the console when dev tools are open:

```
Warning: Received `true` for a non-boolean attribute `monospace`.

If you want to write it to the DOM, pass a string instead: monospace="true" or monospace={value.toString()}.
td
```

Can you have a look at this?

PR #677, opened 2024-12-13
Limit text height in PDF export

weeman1337:
This is not an optimal solution, but the best we can do with pdfmake. If we want to cut off text we need to change pdfmake. For that we need a new ticket with a higher estimation.

With the max height property lines that would not fit into a shape are not printed at all.

| This PR |
| --- |
| image |
| Before |
|  image | 

### DIFF
--- a/packages/react-sdk/src/components/BoardBar/pdf/utils.ts
+++ b/packages/react-sdk/src/components/BoardBar/pdf/utils.ts
@@ -160,6 +160,13 @@ export function textContent(
     lineHeight: 1,
     margin: 0,
     font: 'Inter',
+    // Set maxHeight to prevent text overflows.
+    // This is not optimal, because it does not clip the text.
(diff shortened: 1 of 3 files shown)


packages/react-sdk/src/components/BoardBar/pdf/utils.ts
```diff
@@ -160,6 +160,13 @@ export function textContent(
     lineHeight: 1,
     margin: 0,
     font: 'Inter',
+    // Set maxHeight to prevent text overflows.
+    // This is not optimal, because it does not clip the text.
+    // Instead it does not lines, that would not fit into the shape.
```
weeman1337, line 165:
```suggestion
    // Instead it does not draw lines, that would not fit into the shape.
```

mgcm (approved):
🔧 👍

PR #696, opened 2024-12-27
NEO-954: "no snapshot available" dialog

mgcm:
A blocking information dialog will now be shown when the `net.nordeck.whiteboard.document.create` event is not readable due to a user joining rooms where message history is not available.

This will happen in rooms where history is shared only after invite or join. It also applies to encrypted rooms where message keys are not shared for past events.

Screenshot 2025-01-09 at 15 29 27


.changeset/eleven-pans-pull.md
```diff
@@ -0,0 +1,6 @@
+---
+'@nordeck/matrix-neoboard-widget': patch
+'@nordeck/matrix-neoboard-react-sdk': patch
```
HarHarLinks, line 3: this is a feature!
```suggestion
'@nordeck/matrix-neoboard-widget': minor
'@nordeck/matrix-neoboard-react-sdk': minor
```

packages/react-sdk/src/components/connection-state/ConnectionStateProvider.tsx
```diff
@@ -78,11 +84,24 @@ export const ConnectionStateProvider: React.FC<PropsWithChildren> = function ({
...
+   */
+  useEffect(() => {
+    if (!snapshotInfo.snapshotLoadFailed) {
+      // No load snapshot error - no dialog
+      setSnapshotLoadDialogOpen(false);
+      return;
+    }
+
+    setSnapshotLoadDialogOpen(true);
+  }, [snapshotInfo, snapshotLoadDialogOpen, setSnapshotLoadDialogOpen]);
```
maheichyk, line 98: We can use `snapshotInfo.snapshotLoadFailed` in dependencies to save `useEffect` calls.

packages/react-sdk/src/state/synchronizedDocumentImpl.ts
```diff
@@ -56,6 +58,7 @@ export class SynchronizedDocumentImpl<T extends Record<string, unknown>>
 {
   private readonly logger = getLogger('SynchronizedDocument');
   private readonly loadingSubject = new BehaviorSubject<boolean>(true);
+  private readonly loadingDone = new BehaviorSubject<boolean>(false);
```
maheichyk, line 61: Do we need this second loading subject to describe the behavior? Seems that `loadingSubject` could be just used.

packages/react-sdk/src/components/connection-state/SnapshotLoadStateDialog.tsx
```diff
@@ -0,0 +1,61 @@
...
+  DialogContentText,
+  DialogTitle,
+} from '@mui/material';
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { useSnapshotLoadState } from './useSnapshotLoadState';
+
+export const SnapshotLoadStateDialog: React.FC = () => {
+  const { t } = useTranslation('neoboard');
+  const { snapshotLoadDialogOpen } = useSnapshotLoadState();
```
maheichyk, line 32: I actually wonder if we could use an existing whiteboard instance and use it's state to determine snapshot loading failure. Currently we have only `observeIsLoading` that is not enough but can be changed. In this case we will not need to use store for this data.

maheichyk (changes requested):
Feature looks good to me, I left few questions/comments to the code.

packages/react-sdk/src/state/synchronizedDocumentImpl.ts
mgcm: you are right. removing it and adjusting the tests to no longer depend on the loading status, as it is not relevant for them.

packages/react-sdk/src/components/connection-state/SnapshotLoadStateDialog.tsx
mgcm: I think our discussions have addressed this point and we live with this for now?

maheichyk (approved):

PR #702, opened 2025-01-02
Only show text formatting toolbar buttons when shapes have text

mgcm:
Text formatting tools in the shape toolbar (including text color) will only be shown when applicable, which is when the shape actually has text in it or the user started typing.

**Before**
https://github.com/user-attachments/assets/45e66bb9-4471-422a-8a48-5bafd7f9dbdd

**After**
https://github.com/user-attachments/assets/efc19781-fa44-4c04-80f5-e8903b98cad5

### DIFF
--- a/packages/react-sdk/src/components/ElementBar/ColorPickerButton/TextColorPicker.tsx
+++ b/packages/react-sdk/src/components/ElementBar/ColorPickerButton/TextColorPicker.tsx
@@ -51,6 +51,10 @@ export function TextColorPicker() {
   const defaultShade = hasTextShape ? activeShapeTextShade : activeTextShade;
 
   let color = extractFirstTextColor(activeElements);
+
+  // Gets a default color when there is no text, based on the fill color of the first
+  // selected element. This is required for when the text formatting tools are shown
+  // if you start typing new text into a shape
   if (color == undefined) {
     const element = activeElements[0];
     if (element && 'fillColor' in element) {
--- a/packages/react-sdk/src/components/ElementBar/ColorPickerButton/TextColorPicker.tsx
+++ b/packages/react-sdk/src/components/ElementBar/ColorPickerButton/TextColorPicker.tsx
@@ -17,9 +17,11 @@
 import { useCallback } from 'react';
 import { useTranslation } from 'react-i18next';
 import { findColor, useColorPalette } from '../../../lib';
+import { isEmptyText } from '../../../lib/text-formatting';
 import {
   includesShapeWithText,
   includesTextShape,
+  ShapeElement,
   useActiveElements,
   useElements,
 } from '../../../state';
@@ -44,6 +46,7 @@ export function TextColorPicker() {
   const { activeElementIds } = useActiveElements();
   const activeElementsObject = useElements(activeElementIds);
   const activeElements = Object.values(activeElementsObject);
+  const elements = Object.values(activeElements);
 
   const hasText = includesTextShape(activeElements);
   const hasTextShape = includesShapeWithText(activeElements);
@@ -85,6 +88,12 @@ export function TextColorPicker() {
       setActiveShapeTextShade,
     ],
   );
+
+  if (
+    elements.every((element) => isEmptyText((element as ShapeElement).text))
+  ) {
+    return null;
+  }
 
   return (
     <ColorPicker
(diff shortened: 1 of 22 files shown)


packages/react-sdk/src/components/ElementBar/ColorPickerButton/TextColorPicker.tsx
```diff
@@ -50,7 +50,14 @@ export function TextColorPicker() {
 
   const defaultShade = hasTextShape ? activeShapeTextShade : activeTextShade;
 
-  const color = extractFirstTextColor(activeElements);
+  let color = extractFirstTextColor(activeElements);
+  if (color == undefined) {
```
weeman1337, line 54: Why does it use the fill colour as text colour if there is not text colour?

packages/react-sdk/src/components/ElementBar/ElementBar.tsx
```diff
@@ -25,18 +26,24 @@ import { TextAlignmentButtons } from './TextAlignmentButtons';
...
-      <FontSizeButton />
-      <TextBoldButton />
-      <TextItalicButton />
-      <TextColorPicker />
+      {showTextTools && (
+        <>
+          <FontSizeButton />
+          <TextBoldButton />
+          <TextItalicButton />
+          <TextColorPicker />
```
weeman1337, line 42: I think the initial idea here was to keep the colour buttons next to each other. Can you swap the colour and alignment?

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Selection/types.ts
```diff
@@ -25,4 +25,5 @@ export type WithSelectionProps = {
 export type WithExtendedSelectionProps = WithSelectionProps & {
   activeElementIds?: string[];
   overrides?: Elements;
+  enableTextTools?: (isVisible: boolean) => void;
```
weeman1337, line 28: Maybe I am a bit picky here :D

- `isVisible` should be `enabled`
- Also if you use it, you would expect it to be enabled, if you call the function. Maybe `setTextToolsEnabled` is a better name.

packages/react-sdk/src/components/Whiteboard/WhiteboardHost.tsx
```diff
@@ -63,6 +64,20 @@ const WhiteboardHost = ({
     useLayoutState();
   const { activeElementIds } = useActiveElements();
   const overrides = useElementOverrides(activeElementIds);
+  const [enableTextTools, setEnableTextTools] = useState(false);
+
+  // iterate over all activeElements and check if any of them has text content
+  const hasTextShape = useMemo(() => {
```
weeman1337, line 70: We have a helper function for that already [`includesShapeWithText`](https://github.com/nordeck/matrix-neoboard/blob/4f31dc6162ae992911a8c6faef18bbfb5046949f/packages/react-sdk/src/state/crdt/documents/elements.ts#L284)

weeman1337 (changes requested):
Getting closer :+1: Please have a look at my comments.

Also tested locally. The flickering is gone and it works as expected.

packages/react-sdk/src/components/Whiteboard/WhiteboardHost.tsx
mgcm: Nice! It also needs to use [includesTextShape](https://github.com/nordeck/matrix-neoboard/blob/4f31dc6162ae992911a8c6faef18bbfb5046949f/packages/react-sdk/src/state/crdt/documents/elements.ts#L277) otherwise it will ignore transparent background shapes.

packages/react-sdk/src/components/ElementBar/ElementBar.tsx
mgcm: yup!

packages/react-sdk/src/components/ElementBar/ColorPickerButton/TextColorPicker.tsx
mgcm: if there is no text color set, it will use the best foreground color for text based on the current fill color. That's what [`findForegroundColor`](https://github.com/nordeck/matrix-neoboard/blob/cd8480b8f0b2b5679854b8d84b0d9b885a1cf8ca/packages/react-sdk/src/components/BoardBar/pdf/utils.ts#L188) is doing.

weeman1337: Unfortunately I can't follow. `findForegroundColor` gets a colour as arg. It does not read the fill colour. Maybe let's change the question: Why did you make this change / what happens if we don't change it?

mgcm: added a comment to this as we agreed.

weeman1337 (approved):
Re-tested; Looks good :+1:

PR #706, opened 2025-01-07
Fix MUI dialog issue in fullscreen mode

ahmadkadri:
Problem
When the application enters fullscreen mode, the MUI Dialog component does not render correctly on top of the fullscreen content. This issue occurs because the Dialog by default attaches to document.body, which is outside the fullscreen context (widget-root container). As a result, the dialog either becomes invisible or behaves unexpectedly in fullscreen mode.

solution:
Ensure the Dialog renders properly within the fullscreen to explicitly use the container prop of the MUI Dialog, anchoring it to the widget-root container.

Before
https://github.com/user-attachments/assets/661f06df-7173-4671-a52e-cd5035fab4b4

After
https://github.com/user-attachments/assets/5930d195-c36c-47e3-ac34-02ea883fd3c6

### DIFF
--- a/packages/react-sdk/src/components/BoardBar/ExportWhiteboardDialog.tsx
+++ b/packages/react-sdk/src/components/BoardBar/ExportWhiteboardDialog.tsx
@@ -18,7 +18,6 @@ import CloseIcon from '@mui/icons-material/Close';
 import {
   Alert,
   Button,
-  Dialog,
   DialogActions,
   DialogContent,
   DialogContentText,
@@ -35,6 +34,7 @@ import {
 import { unstable_useId as useId, visuallyHidden } from '@mui/utils';
 import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
+import { Dialog } from '../common/Dialog';
 import { ExportWhiteboardDialogDownloadFile } from './ExportWhiteboardDialogDownloadFile';
 import { ExportWhiteboardDialogDownloadPdf } from './ExportWhiteboardDialogDownloadPdf';
 
@@ -54,7 +54,6 @@ export function ExportWhiteboardDialog({
     <Dialog
       aria-labelledby={dialogTitleId}
       aria-describedby={dialogDescriptionId}
-      container={document.getElementById('widget-root')}
       open={open}
       onClose={onClose}
       maxWidth="sm"
--- a/packages/react-sdk/src/components/BoardBar/ExportWhiteboardDialog.tsx
+++ b/packages/react-sdk/src/components/BoardBar/ExportWhiteboardDialog.tsx
@@ -54,6 +54,7 @@ export function ExportWhiteboardDialog({
     <Dialog
       aria-labelledby={dialogTitleId}
       aria-describedby={dialogDescriptionId}
+      container={document.getElementById('widget-root')}
       open={open}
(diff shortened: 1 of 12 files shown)


packages/react-sdk/src/components/Layout/useFullscreenMode.ts
```diff
@@ -55,10 +55,8 @@ export function useFullscreenMode(): UseFullscreenModeResult {
     isFullscreenRequestRunning.current = true;
 
     if (value && !document.fullscreenElement) {
-      const container =
-        document.getElementById('widget-root') ?? document.documentElement;
       try {
-        await container.requestFullscreen();
+        await document.body.requestFullscreen();
```
weeman1337, line 59: This change breaks fullscreen in NeoBoard standalone:

image

It should only show the canvas full screen.

weeman1337 (changes requested):
Unfortunately this change would break standalone

packages/react-sdk/src/components/BoardBar/ExportWhiteboardDialog.tsx
```diff
@@ -54,6 +54,7 @@ export function ExportWhiteboardDialog({
     <Dialog
       aria-labelledby={dialogTitleId}
       aria-describedby={dialogDescriptionId}
+      container={document.getElementById('widget-root')}
```
weeman1337, line 57: If we create a new dialog we will have forgotten that…

What we can do is create a new Dialog Component in NeoBoard, SDK that has a default value for the container. Then use this Dialog instead. We can also add an [ESLint rule](https://github.com/nordeck/matrix-neoboard/blob/1d3a38ab604584369389e74cd8da0fd9c5c0a913/eslint.config.mjs#L69) to prevent people from using the plain MUI dialog.

packages/react-sdk/src/components/DeveloperTools/DeveloperToolsDialog.tsx
```diff
@@ -44,7 +44,7 @@ export function DeveloperToolsDialog({
 }) {
   const { t } = useTranslation('neoboard');
 
-  const { document, communicationChannel } =
+  const { document: communicationDocument, communicationChannel } =
```
weeman1337, line 47: Is there a reason why you renamed `document` here?

weeman1337 (changes requested):
I think we should introduce a new Dialog component. See my comment.

packages/react-sdk/src/components/BoardBar/ExportWhiteboardDialog.tsx
MTRNord: It should be in toolkit imho just like the rest of this https://github.com/nordeck/matrix-widget-toolkit/blob/main/packages/mui/src/components/MuiThemeProvider/theme.ts#L246-L251

weeman1337: So with `document.fullscreenElement ?? document.documentElement` this works without using some IDs. Tested and works. I would say we should then just set the default prop in the toolkit as MTRNord mentioned.

If you want we can have a call about that.

Sorry for the extra round Ahmad :see_no_evil:

PR #722, opened 2025-01-17
Checks for image build+push step success and required output before SBOM scan

mgcm:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,6 +100,7 @@ jobs:
             REACT_APP_REVISION=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.revision'] }}
 
       - name: Run Trivy to get an SBOM report of the container
+        if: ${{ success() && steps.build_and_push.outputs.digest }}
         env:
           IMAGE_REF: ${{ env.DOCKER_IMAGE }}@${{ steps.build_and_push.outputs.digest }}
         uses: aquasecurity/trivy-action@0.29.0
@@ -112,6 +113,7 @@ jobs:
           github-pat: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Upload SBOM report as a build artifact
+        if: ${{ success() && steps.build_and_push.outputs.digest }}
         uses: actions/upload-artifact@v4
         with:
           name: sbom-spdx-report
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,7 +93,6 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
           platforms: linux/amd64,linux/arm64,linux/s390x
           sbom: true
-          outputs: 'type=image'
           provenance: true
           build-args: |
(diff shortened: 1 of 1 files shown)


HarHarLinks:
Any links to github actions docs for why this works?

mgcm:
> Any links to github actions docs for why this works?

`Note that in conditionals, falsy values (false, 0, -0, "", '', null) are coerced to false and truthy (true and other non-falsy values) are coerced to true.` in https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions

HarHarLinks:
Okay, that is one, but it would have saved me time not having to find the docs for `success` and the build-and-push action's outputs myself ;)

HarHarLinks (commented):
> success()
Returns `true` when all previous steps have succeeded.

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#success

For `steps.build_and_push.outputs.digest`:

https://github.com/docker/build-push-action?tab=readme-ov-file#outputs, does what you'd expect. However it does not explain when this output does (not) exist.

I would prefer to understand: Is not either condition enough?

mgcm:
> > success()
> > Returns `true` when all previous steps have succeeded.
> 
> https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#success
> 
> For `steps.build_and_push.outputs.digest`:
> 
> https://github.com/docker/build-push-action?tab=readme-ov-file#outputs, does what you'd expect. However it does not explain when this output does (not) exist.
> 
> I would prefer to understand: Is not either condition enough?

`success()` on it's own would not stand, as the `build_and_push` action will succeed but might not have pushed a container onto the registry. you can argue that it might not be required at all, though.

As for the output digest, it will be an empty string if the `push` option is false and thus evaluate to falsy.

mgcm:
closing in favor of #725

mgcm:
Reopening because #725 has the issue that images built using `buildx` are not available in the local docker registry:

> Buildx has some nice new features like support for better caching between images as well as cleaner output during builds. However, it runs completely independently of your standard local docker registry. If you run docker ps, you’ll see a buildx builder running as a container on your local machine. This is a virtual builder that we created using docker buildx create. There are ways to create builders that run as Kubernetes pods or remote Docker machines as well but these require a lot more setup. **Unfortunately because of this there’s no way to manipulate the images created with buildx after the fact.** You can no longer do docker push since the image doesn’t exist in your registry so you must add docker buildx build --push to your build command which will push to a remote registry right after the build is finished.

- in https://speedscale.com/blog/how-to-build-multi-arch-docker-images/

One solution would be to use `type=docker` but then [this doesn't support multi-platform builds](https://docs.docker.com/reference/cli/docker/buildx/build/#docker).

One could also export the images as tarballs and scan the exported OCI files in these cases but this adds even more complexity to the workflow.

I suggest we rollback the changes #725 introduced and use the approach in this PR instead.

ahmadkadri (approved):
LGTM

PR #742, opened 2025-02-07
Font is Undefined in Old Board Snapshots

ahmadkadri:
In old board snapshots, selecting text shows no default font value. If the font family is undefined, it should default to "Inter."

### DIFF
--- a/packages/react-sdk/src/components/ElementBar/FontFamilyButton/FontFamilyButton.tsx
+++ b/packages/react-sdk/src/components/ElementBar/FontFamilyButton/FontFamilyButton.tsx
@@ -50,7 +50,7 @@ export function FontFamilyButton() {
       size="small"
       variant="standard"
       disableUnderline={true}
-      value={activeFontFamily ?? 'Inter'}
+      value={activeFontFamily}
       inputProps={{
         'aria-label': t('elementBar.fontForm', 'Select font family'),
       }}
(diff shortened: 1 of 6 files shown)


packages/react-sdk/src/components/ElementBar/FontFamilyButton/FontFamilyButton.tsx
```diff
@@ -50,7 +50,7 @@ export function FontFamilyButton() {
       size="small"
       variant="standard"
       disableUnderline={true}
-      value={activeFontFamily}
+      value={activeFontFamily ?? 'Inter'}
```
weeman1337, line 53: I don't really like this. The type promises, that there is always a font family set. See [here](https://github.com/nordeck/matrix-neoboard/blob/535f49872de98979cfca71245c134f96028808f3/packages/react-sdk/src/components/Layout/useLayoutState.tsx#L117).

We should fix this at the place where we set the active font family. There should never be an undefined value in the layout state in the first place.

weeman1337 (changes requested):

packages/react-sdk/src/components/ElementBar/FontFamilyButton/FontFamilyButton.tsx
weeman1337: This has been done

PR #753, opened 2025-02-18
Initial matrixRTC implementation

mgcm:

- [X] Added or updated documentation.


packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -0,0 +1,386 @@
...
+  }
+
+  destroy(): void {
+    this.destroySubject.next();
+    this.sessionJoinedSubject.complete();
+    this.sessionSubject.complete();
+    this.sessionLeftSubject.complete();
+  }
+
+  private handleRTCSessionEvent(
```
maheichyk, line 155: I think it should be possible to simplify code that handles updates and additions of events, maybe some more tests could be added.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -0,0 +1,386 @@
...
+    event: Partial<StateEvent<RTCSessionEventContent>>,
+  ): void {
+    this.logger.debug(`Session ${event.state_key} left whiteboard`);
+
+    this.sessions = this.sessions.filter(
+      (s) => s.session_id !== event.state_key,
+    );
+    this.sessionLeftSubject.next({
+      sessionId: event.state_key!,
+      userId: event.sender!,
```
maheichyk, line 233: suggest to import types to not use `!` operator

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -0,0 +1,386 @@
...
+        sessionId: baseSession.session_id,
+        ...baseSession, // Include any additional fields
+      };
+
+      // Apply the patch function to get updated values
+      const patchedSession = patchFn(sessionForPatch);
+
+      // Merge the patched fields back into the base session
+      // This ensures all fields from the patch are applied
+      const updatedSession = {
```
maheichyk, line 349: maybe can import this part, looks complicated

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -0,0 +1,404 @@
...
+  BehaviorSubject,
+  distinctUntilChanged,
+  mergeMap,
+  Observable,
+  Subject,
+  switchMap,
+  takeUntil,
+} from 'rxjs';
+import { PeerConnection } from './connection';
+import { MatrixRtcPeerConnection } from './connection/matrixRtcPeerConnection';
```
maheichyk, line 31:
```suggestion
import { MatrixRtcPeerConnection } from './connection';
```

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -0,0 +1,404 @@
...
+  PeerConnectionStatistics,
+} from './types';
+import { observeVisibilityState } from './visibilityState';
+
+export interface SFUConfig {
+  url: string;
+  jwt: string;
+}
+
+export function sfuConfigEquals(a?: SFUConfig, b?: SFUConfig): boolean {
```
maheichyk, line 47: function not used

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -0,0 +1,404 @@
...
+  if (a === undefined && b === undefined) return true;
+  if (a === undefined || b === undefined) return false;
+
+  return a.jwt === b.jwt && a.url === b.url;
+}
+
+export class MatrixRtcCommunicationChannel implements CommunicationChannel {
```
maheichyk, line 54: `MatrixRtcCommunicationChannel` seems to be not too much different from `WebRtcCommunicationChannel`, maybe could use some abstraction to reuse the code and make possible to have other setups of rtc, livekit if would make sense.

packages/react-sdk/src/state/communication/signaling/toDeviceMessageSignalingChannel.test.ts
```diff
@@ -46,8 +45,20 @@ describe('ToDeviceMessageSignalingChannel', () => {
...
         '@connection-id',
         [
-          new RTCIceCandidate({ candidate: 'candidate-0' }),
-          new RTCIceCandidate({ candidate: 'candidate-1' }),
+          new RTCIceCandidate({
+            candidate:
+              'candidate:702786350 2 udp 41819902 8.8.8.8 60769 typ relay raddr 8.8.8.8',
+            sdpMLineIndex: null,
+            sdpMid: null,
+            usernameFragment: null,
```
maheichyk, line 53: please add some comment why we have changed that (something about livekit)

packages/react-sdk/src/state/useActiveWhiteboardMembers.ts
```diff
@@ -24,20 +28,35 @@ export type ActiveWhiteboardMember = {
...
+      Object.values(statistics.communicationChannel.peerConnections).forEach(
+        (p) => {
+          if (isPeerConnected(p)) {
+            activeWhiteboardMembers.set(p.remoteUserId, {
+              userId: p.remoteUserId,
+            });
+          }
+        },
+      );
+      return Array.from(activeWhiteboardMembers.values());
```
maheichyk, line 59: I think this like with Array creation is kind of duplicated for other if branch. I suggest to move it after the if block.

docs/adrs/adr009-matrix-rtc-livekit.md
```diff
@@ -0,0 +1,86 @@
+# ADR009: MatrixRTC + LiveKit for Real Time Communications
+
+Status: draft
```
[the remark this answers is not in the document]
mgcm: yes, just updated following the more traditional structure.

matrix-neoboard-widget/src/widgetCapabilities.ts
```diff
@@ -110,8 +102,35 @@ export const widgetCapabilities = [
...
+  widgetCapabilities.push(
+    WidgetEventCapability.forStateEvent(
+      EventDirection.Send,
+      STATE_EVENT_RTC_MEMBER,
+      // We only need to write the own state, but read state from everyone
+      `_${userId}_${deviceId}`,
+    ),
+    WidgetEventCapability.forStateEvent(
+      EventDirection.Receive,
+      STATE_EVENT_RTC_MEMBER,
```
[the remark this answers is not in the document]
mgcm: I made a note of this in the Consequences section of the ADR.

packages/react-sdk/src/model/matrixRtcSessions.ts
```diff
@@ -0,0 +1,104 @@
...
+import Joi from 'joi';
+import { RTCFocus } from '../../src/state/communication/discovery';
+import { isValidEvent } from './validation';
+
+export const DEFAULT_RTC_EXPIRE_DURATION = 1000 * 60 * 60 * 4;
+
+// unstable prefix for m.rtc.member in MSC4143
+export const STATE_EVENT_RTC_MEMBER = 'org.matrix.msc3401.call.member';
+
+export type RTCSessionEventContent = {
```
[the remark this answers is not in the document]
mgcm: Added a reference to the CallMembership definition in the JS SDK, which has good documentation for each property.

packages/react-sdk/src/model/matrixRtcSessions.ts
```diff
@@ -0,0 +1,104 @@
...
+  call_id: string;
+  scope: string | undefined;
+  application: string;
+  session_id: string;
+  whiteboard_id: string;
+  user_id: string;
+  device_id: string;
+  createdTs: number | undefined;
+  expires: number | undefined;
```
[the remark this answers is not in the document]
mgcm: for these two specifically, I am following the CallMembership definition, which has them potentially undefined. But I'm open to change this.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -0,0 +1,386 @@
...
+  RTCSessionEventContent,
+  STATE_EVENT_RTC_MEMBER,
+  isRTCSessionNotExpired,
+  isValidRTCSessionStateEvent,
+  newRTCSession,
+} from '../../../model';
+import { DEFAULT_RTC_EXPIRE_DURATION } from '../../../model/matrixRtcSessions';
+import { Session, SessionManager } from './types';
+
+export type SessionState = {
```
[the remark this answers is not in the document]
mgcm: yup, removed this.

packages/react-sdk/src/state/communication/signaling/toDeviceMessageSignalingChannel.test.ts
mgcm: When adding the `livekit-client` package, it brings in the [webrtc-adapter](https://github.com/webrtchacks/adapter) package which adds a shim to RTCIceCandidate which uses the [sdp](https://github.com/otalk/sdp) packege to parse the actual string content and it fails with the previous `candidate-0` value, which is not a  valid SDP format.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
mgcm: I see your point but I think that such effort can be done later when / if the need arises.

docs/adrs/adr009-matrix-rtc-livekit.md
```diff
@@ -0,0 +1,123 @@
...
+  "event_id": "$bFsA4Obl-sneiJlq4SAM2WGMLe00ie3f-Mod7VQfF_c",
+  "room_id": "!BWCjlIjHYWgJyZySxE:matrix.internal"
+}
+```
+
+This state event is kept in sync with the realtime LiveKit connection status, having it's `content` cleared when that realtime connection is lost (either intentionally or not).
+
+### Signaling
+
+Thanks to the LiveKit [Client JS SDK][livekit-js-sdk], we don't have to handle estabilishing WebRTC peer connections to every participant. This is now done by the SDK itself, abstracted away by having a [server-side room](livekit-room) to which each participant connects to. There is still a WebRTC negotiation process that establishes the two data connections to the LiveKit backend but that is completely opaque and we only have to monitor a single connection status.
```
maheichyk, line 66:
```suggestion
Thanks to the LiveKit [Client JS SDK][livekit-js-sdk], we don't have to handle establishing WebRTC peer connections to every participant. This is now done by the SDK itself, abstracted away by having a [server-side room](livekit-room) to which each participant connects to. There is still a WebRTC negotiation process that establishes the two data connections to the LiveKit backend but that is completely opaque and we only have to monitor a single connection status.
```

docs/adrs/adr009-matrix-rtc-livekit.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Deployment
+
+Two new backend services are required: the LiveKit Server and the LiveKit JWT Service. This increases the complexity of deploying the widget but as these components are also a requirement for Element Call, we are positive that they will become a standard and will be available on most Matrix deployments.
+
+### Multiple RTC apps
+
+The currently proposed specification allows for having RTC membership state events for different applications, by using the `application` property. For example, this is `m.call` for Element Call and `net.nordeck.whiteboard` for NeoBoard.
+
+This is fine if you are using a single RTC app within a matrix room but there is currently no specification of the expected behaviour if a user adds more than one MatrixRTC-based application to the same room and tries to use them at the same time. They will be racing to update the same membership state event, which is obviosuly not desirable.
```
maheichyk, line 88:
```suggestion
This is fine if you are using a single RTC app within a matrix room but there is currently no specification of the expected behaviour if a user adds more than one MatrixRTC-based application to the same room and tries to use them at the same time. They will be racing to update the same membership state event, which is obviously not desirable.
```

packages/react-sdk/src/model/matrixRtcSessions.ts
```diff
@@ -0,0 +1,107 @@
...
+import Joi from 'joi';
+import { RTCFocus } from '../state/communication';
+import { isValidEvent } from './validation';
+
+export const DEFAULT_RTC_EXPIRE_DURATION = 1000 * 60 * 60 * 4;
+
+// unstable prefix for m.rtc.member in MSC4143
+export const STATE_EVENT_RTC_MEMBER = 'org.matrix.msc3401.call.member';
+
+// Folowing Matrix JS SDK's SessionMembershipData
```
maheichyk, line 27:
```suggestion
// Following Matrix JS SDK's SessionMembershipData
```

packages/react-sdk/src/model/matrixRtcSessions.ts
```diff
@@ -0,0 +1,107 @@
...
+// see https://github.com/matrix-org/matrix-js-sdk/blob/d6ede767c929f7be179d456b5a0433be21ccaf7c/src/matrixrtc/CallMembership.ts#L35
+export type RTCSessionEventContent = {
+  application: string;
+  call_id: string;
+  device_id: string;
+  focus_active: RTCFocus;
+  foci_preferred: RTCFocus[];
+  created_ts: number | undefined;
+  scope: string | undefined;
+  expires: number | undefined;
```
maheichyk, line 37:
```suggestion
  created_ts?: number;
  scope?: string;
  expires?: number;
```
Similar to other places and widgets, easier to read.

packages/react-sdk/src/model/matrixRtcSessions.ts
```diff
@@ -0,0 +1,107 @@
...
+  Joi.object().max(0),
+
+  // Full object option with all the requirements
+  Joi.object<RTCSessionEventContent, true>({
+    application: Joi.string().required(),
+    call_id: Joi.string().required(),
+    device_id: Joi.string().required(),
+    focus_active: Joi.object().required(),
+    foci_preferred: Joi.array().items(Joi.object()).required(),
+    created_ts: Joi.number().optional(),
```
maheichyk, line 55: I don't think we need `optional` it is optional by default

packages/react-sdk/src/model/matrixRtcSessions.ts
```diff
@@ -0,0 +1,107 @@
...
+    true,
+  );
+}
+
+export function isWhiteboardRTCSessionStateEvent(
+  event: StateEvent<unknown>,
+): event is StateEvent<RTCSessionEventContent> {
+  return (
+    isValidRTCSessionStateEvent(event) &&
+    (event.content.application === 'net.nordeck.whiteboard' ||
```
maheichyk, line 77: maybe make `net.nordeck.whiteboard` a constant, there is one more place where it is used

packages/react-sdk/src/model/matrixRtcSessions.ts
```diff
@@ -0,0 +1,107 @@
...
+      Object.keys(event.content).length === 0)
+  );
+}
+
+export function isRTCSessionNotExpired(
+  event: StateEvent<RTCSessionEventContent>,
+): boolean {
+  if (!event.content || Object.keys(event.content).length === 0) {
+    return false;
+  }
```
maheichyk, line 87: I think we don't need this content validation. The event is already validated when this function is used as far as I can see.

packages/react-sdk/src/model/matrixRtcSessions.ts
```diff
@@ -0,0 +1,107 @@
...
+    event.content.expires !== undefined && event.content.expires > Date.now()
+  );
+}
+
+export function newRTCSession(
+  deviceId: string,
+  whiteboardId: string,
+): RTCSessionEventContent {
+  return {
+    application: 'net.nordeck.whiteboard',
```
maheichyk, line 98: suggest to use constant

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -0,0 +1,396 @@
...
+  private async connect() {
+    if (this.statistics.localSessionId) {
+      this.logger.log('Communication channel is already open');
+      return;
+    }
+
+    this.logger.log('Connecting communication channel');
+    const { sessionId } = await this.sessionManager.join(this.whiteboardId);
+
+    const widgetApi = await Promise.resolve(this.widgetApiPromise);
```
maheichyk, line 182:
```suggestion
    const widgetApi = await this.widgetApiPromise;
```

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -0,0 +1,396 @@
...
+
+  private async initLiveKitServer(session: Session): Promise<void> {
+    if (this.sfuConfig !== undefined) {
+      this.logger.debug('LiveKit config already set, skipping init');
+      // LiveKit servers already known
+      return;
+    }
+
+    try {
+      const widgetApi = await Promise.resolve(this.widgetApiPromise);
```
maheichyk, line 209:
```suggestion
      const widgetApi = await this.widgetApiPromise;
```

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -0,0 +1,314 @@
...
+  newRTCSession,
+} from '../../../model';
+import {
+  DEFAULT_RTC_EXPIRE_DURATION,
+  isWhiteboardRTCSessionStateEvent,
+} from '../../../model/matrixRtcSessions';
+import { SessionState } from './sessionManagerImpl';
+import { Session, SessionManager } from './types';
+
+export class RTCSessionManagerImpl implements SessionManager {
```
maheichyk, line 42: Suggest to rename it to `MatrixRtcSessionManagerImpl` to have consistent name with file and other code.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -0,0 +1,396 @@
...
+        session.sessionId,
+      );
+      const peerConnection = new MatrixRtcPeerConnection(
+        session,
+        this.sfuConfig,
+      );
+      this.peerConnections.push(peerConnection);
+
+      peerConnection.observeMessages().subscribe((m) => {
+        return this.messagesSubject.next(m);
```
maheichyk, line 253: I observer messages when users are on different boards in different rooms. For example cursors, could please check?

maheichyk (changes requested):
Thanks for the changes and ADR! I have left few comments, please have a look.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
```diff
@@ -0,0 +1,298 @@
...
+      }),
+    );
+
+    await expect(leftPromise).resolves.toEqual([
+      { sessionId: '_@another-user_ANOTHERDEVICEID', userId: '@user-id' },
+    ]);
+  });
+
+  it('should leave a whiteboard', async () => {
+    widgetApi.mockSendStateEvent(mockWhiteboardSessions());
```
maheichyk, line 263: I think we don't need this line

packages/react-sdk/src/model/matrixRtcSessions.ts
mgcm: I thought about that but then you get:

```
application: STATE_EVENT_WHITEBOARD,
```

which I think is semantically misleading. 

I would rather define a new constant instead.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
mgcm: Yes, there was an issue with the `roomName` parameter which is part of the payload to get the JWT token. It had a default value that I forgot to override. Solved that, great catch.

docs/adrs/adr009-matrix-rtc-livekit.md
```diff
@@ -0,0 +1,123 @@
+# ADR009: MatrixRTC + LiveKit for Real Time Communication
+
+Status: draft
+
+## Context
+
+This ADR supercedes and deprecates [ADR006][adr006] because of the scalability and stability issues we encountered when using the peer-to-peer full-mesh approach.
+
+When MatrixRTC was evaluated previously, there were a number of issues that have since been resolved. Widgets now [have access to the user's own device id](widget-api-device-id) and since the introduction of initial support for [MSC4143](MSC4143) in Element Web, via the Group Calls feature, data-only calls are no longer displayed in the timeline.
```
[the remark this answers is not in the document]
mgcm: Not sure what you mean by suggestions. Can you give an example?

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
mgcm: Indeed, the join is already sending a state event if required, so this is not necessary.

mgcm:
> Should this PR add something similar to https://github.com/nordeck/matrix-neoboard/blob/main/docs/model/webrtc-messages.md?

@HarHarLinks The new ADR mentions the only model change, which is the replacement of `net.nordeck.whiteboard.sessions` with  `org.matrix.msc3401.call.member`

docs/adrs/adr009-matrix-rtc-livekit.md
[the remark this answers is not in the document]
maheichyk: I meant to have fewer maximum characters per line, but I see you have already changed that 👍

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -0,0 +1,398 @@
...
+      this.logger.debug('LiveKit config already set, skipping init');
+      // LiveKit servers already known
+      return;
+    }
+
+    try {
+      const widgetApi = await this.widgetApiPromise;
+
+      const focus = (
+        await makePreferredLivekitFoci(session, widgetApi.widgetId)
```
maheichyk, line 212: Should we use a room id here instead `widgetId`? In some special setup we could have several neoboard widgets in one room. I think this is related to https://github.com/nordeck/matrix-neoboard/pull/753#discussion_r2038333306

packages/react-sdk/src/model/matrixRtcSessions.ts
```diff
@@ -0,0 +1,109 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { StateEvent } from '@matrix-widget-toolkit/api';
+import Joi from 'joi';
+import { RTCFocus } from '../state/communication';
+import { isValidEvent } from './validation';
+
+export const RTC_WHITEBOARD_APPID = 'net.nordeck.whiteboard.application';
```
maheichyk, line 22:
```suggestion
export const RTC_APPLICATION_WHITEBOARD = 'net.nordeck.whiteboard';
```
In my opinion it is fine to have a shorter value that you have defined before (as the sate event we have already).

packages/react-sdk/src/model/matrixRtcSessions.ts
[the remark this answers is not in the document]
maheichyk: For `created_ts` we don't need it from out data model perspective. However the MSC4143 draft spec says:

```
created_ts - timestamp in milliseconds since UNIX epoch.

    this should not be present the first time that the m.rtc.member event is sent.
    if the m.rtc.member event is sent again, the created_ts should be populated with the origin_server_ts that was given to the previous version of the state event.
```

For `expires` we could keep it undefined, even though we always set it, but that is kind of fallback if delayed events do not work.

HarHarLinks:
> > Should this PR add something similar to https://github.com/nordeck/matrix-neoboard/blob/main/docs/model/webrtc-messages.md?
> 
> @HarHarLinks The new ADR mentions the only model change, which is the replacement of `net.nordeck.whiteboard.sessions` with `org.matrix.msc3401.call.member`

@mgcm Ah yes I see webrtc-messages.md is about something else, but would then https://github.com/nordeck/matrix-neoboard/blob/main/docs/model/matrix-events.md be appropriate place to mention this replacement for MatrixRTC mode?

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
mgcm: This is a tough one. Using `widgetId` here seems more appropriate to me, as you can have multiple widgets in the same room and I can see a point to not having all the users in all the widgets sharing the same data pool. Also, the `widgetId` is the `whiteboard-id`. Our data model does support multiple whiteboards in one room but we are ignoring that and just loading the existing whiteboard when instancing new widgets, which leads to an intersting situation where the whiteboard id doesn't match the widget id... Let's discuss this synchronously.

packages/react-sdk/src/model/matrixRtcSessions.ts
mgcm: We need to also process rtc membership events with empty content, which indicate the end of a user's session and update the UI (ie, the collaborators list) when that happens, so they must not be filtered out.

packages/react-sdk/src/model/matrixRtcSessions.ts
mgcm: It was there mostly for readibiity but we haven't really used it in other schemas, so ok.

packages/react-sdk/src/model/matrixRtcSessions.ts
mgcm: after our discussion that removes `created_ts` and makes `expires` required, I also considered that `scope` should be required, as it's not relevant for our app right now but could be in a future where we support for than one board session per-room, so this change doesn't apply any longer as there are no optional properties in the membership state event content.

packages/react-sdk/src/model/matrixRtcSessions.ts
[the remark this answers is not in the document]
mgcm: After our discussion, I removed `created_ts` and made `expires` required.

mgcm:
@HarHarLinks I have added a `matrix-rtc-events.md` document in the model folder that goes into the details of the session changes when using MatrixRTC. I think this covers what you pointed out.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -0,0 +1,412 @@
...
+      this.sfuConfig = await getSFUConfigWithOpenID(widgetApi, focus);
+
+      this.logger.debug('Got SFU config', JSON.stringify(this.sfuConfig));
+
+      if (!this.sfuConfig) {
+        this.logger.warn('Failed to get LiveKit JWT');
+        throw new Error('Failed to get LiveKit JWT');
+      }
+    } catch (error) {
+      this.logger.error('LiveKit SFU setup timed out', error);
```
maheichyk, line 237: This error could happen in more general case, when widget parameter is missing for example. Could we make this error more generic? 'Could not setup LiveKit SFU' for example.

packages/react-sdk/src/model/matrixRtcSessions.test.ts
```diff
@@ -0,0 +1,100 @@
...
+  beforeEach(() => {
+    vi.spyOn(Date, 'now').mockImplementation(() => fixedDate);
+  });
+
+  afterEach(() => {
+    vi.spyOn(Date, 'now').mockRestore();
+  });
+
+  it('should accept event', () => {
+    expect(isValidRTCSessionStateEvent(mockWhiteboardMembership())).toBe(true);
```
maheichyk, line 40: It is more common to pass explicitly the event to validate and not to use a mock function. That would be similar to other tests in this file and to `whiteboardSessions.test.ts` and some other places as well. I think you did that originally and reverted in last commit.

packages/react-sdk/src/components/Whiteboard/constants.ts
```diff
@@ -19,6 +19,8 @@ import { getEnvironment } from '@matrix-widget-toolkit/mui';
 export const infiniteCanvasMode =
   getEnvironment('REACT_APP_INFINITE_CANVAS') === 'true';
 
+export const matrixRtcMode = getEnvironment('REACT_APP_RTC') === 'matrixrtc';
```
maheichyk, line 22: We should update './index.ts' file to reference `matrixRtcMode` constant and update imports to be similar to: `import { matrixRtcMode } from '../../Whiteboard';`

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
mgcm: Yes, I've updated it to be more generic.

packages/react-sdk/src/components/Whiteboard/constants.ts
mgcm: Also done!

packages/react-sdk/src/model/matrixRtcSessions.test.ts
mgcm: I would rather keep it like this. The mocked membership is supposed to be a valid event and I've updated it to accept either a partial content that gets merged to the default values or just replaces it with the object that is passed in. I think this way its more clear what properties are being changed and how this should affect the tests.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -0,0 +1,418 @@
...
+      throw new Error('Unknown session id on session join');
+    }
+
+    // Wait for the Widget API and LiveKit servers to be ready
+    await this.widgetApiPromise;
+    await this.initLiveKitServer(session);
+
+    // only start peer connection for the current session
+    // because matrix rtc is not peer to peer
+    if (this.sfuConfig && session.sessionId == sessionId) {
```
maheichyk, line 257:
```suggestion
    if (this.sfuConfig && session.sessionId === sessionId) {
```

maheichyk (approved):
🚀

PR #771, opened 2025-03-10
Add first version of frames

weeman1337:


packages/react-sdk/src/components/ElementBar/DuplicateActiveElementButton/DuplicateActiveElementButton.test.tsx
```diff
@@ -290,3 +302,41 @@ describe('duplicate', () => {
     });
   });
 });
+
+describe('<DuplicateActiveElementButton />', () => {
+  let widgetApi: MockedWidgetApi;
+  let Wrapper: ComponentType<PropsWithChildren<{}>>;
```
jaller94, line 308: `deno lint` always tells me that `{}` isn't what I expect it to be.
I should use `Record<never, never>`.

jaller94 (changes requested):

docs/model/crdt-documents.md
```diff
@@ -261,6 +267,32 @@ An image element
 }
 ```
 
+### `FrameElement`
+
+(Only with feature flag `REACT_APP_INFINITE_CANVAS`).
+
+A frame element. Used to group elements and run structured presentations.
```
jaller94, line 274:
```suggestion
A frame element is used to group elements and run structured presentations.
```

packages/react-sdk/src/components/ToolsBar/ToolsBar.test.tsx
```diff
@@ -193,4 +190,24 @@ describe('<ToolsBar/>', () => {
     expect(penTool).toBeChecked();
     expect(slide?.getActiveElementIds()).toEqual([]);
   });
+
+  it('should not show the create frame button per default', () => {
```
jaller94, line 194:
```suggestion
  it('should not show the create frame button by default', () => {
```

packages/react-sdk/src/components/ElementBar/DuplicateActiveElementButton/DuplicateActiveElementButton.test.tsx
weeman1337: We don't use Deno. Our ESLint config has this excluded https://github.com/nordeck/matrix-neoboard/blob/main/eslint.config.mjs#L99

jaller94: I know we're not using Deno. Just pointing out that their linter taught me to not use `{}` anymore, because
> {} represents any non-nullish value, including literals like 0 and ""

weeman1337: I do not disagree. That is how we have it currently coded into our linter rules. Can you please bring this to our daily discussion? We need to create a ticket and change the ESLint rules.

To bring this forward: type specified here.

packages/react-sdk/src/components/ElementBar/ColorPickerButton/ElementColorPicker.tsx
```diff
@@ -45,7 +45,9 @@ export function ElementColorPicker() {
 
   const hasShape = activeElements.some((e) => e.type === 'shape');
   const hasOther = activeElements.some((e) => e.type !== 'shape');
-  const hasOnlyImages = activeElements.every((e) => e.type === 'image');
+  const hasOnlyNonColoredElements = activeElements.every(
+    (e) => e.type !== 'shape' && e.type !== 'path',
+  );
```
jaller94, line 50: Why was this changed from images to non-colored elements?

packages/react-sdk/src/components/ToolsBar/ToolsBar.test.tsx
```diff
@@ -193,4 +190,24 @@ describe('<ToolsBar/>', () => {
...
+
+  it('should not show the create frame button by default', () => {
+    render(<ToolsBar />, { wrapper: Wrapper });
+
+    expect(
+      screen.queryByRole('button', { name: 'Create frame' }),
+    ).not.toBeInTheDocument();
+  });
+
+  it('should show the create frame button, if infinite canvas is enabled', () => {
```
jaller94, line 202:
```suggestion
  it('should show the create frame button, if Infinite Canvas is enabled', () => {
```

packages/react-sdk/src/lib/filterRecord.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+export const filterRecord = <V>(
+  items: Record<string, V>,
+  filterFunc: (item: V) => boolean,
+) => {
+  const filtered: Record<string, V> = {};
+
+  Object.entries(items).forEach(([key, item]) => {
+    if (filterFunc(item)) {
+      filtered[key] = item;
```
jaller94, line 32: If `key` is user input, this is may cause object prototype pollution.
I would add them all to a `Map` and let JavaScript convert the map to an Object.

jaller94 (changes requested):
Looks mostly good, but I've got a question and a potential vulnerability.

packages/react-sdk/src/lib/filterRecord.ts
weeman1337: Is this really an issue for the filter function? It iterates an existing object and assigns the same keys. So if there is something bad, then it was already there before. Otherwise please explain how this can go wrong.

packages/react-sdk/src/components/ElementBar/ColorPickerButton/ElementColorPicker.tsx
weeman1337: This is only used here https://github.com/nordeck/matrix-neoboard/pull/771/files#diff-16237f46fd34aa510aa49dc62dcedba6050303bacc275251c5523bf2d36e3e4fR93 . If there is nothing selected, that has a colour, then do nothing. So instead of `hasOnlyImagesOrFramesOrWhateverComesInTheFuture` I think that what we want to have here is without colour.

maheichyk (commented):
From code looks good to me. Tested, seems like resizing of frame does not work. Could you check please?

maheichyk (changes requested):

maheichyk (approved):

packages/react-sdk/src/lib/filterRecord.ts
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @param items - items to filter
+ * @param filterFunc - filter function
+ * @returns record with all props for which the filter function returned true
+ */
+export const filterRecord = <V>(
+  items: Record<string, V>,
+  filterFunc: (item: V) => boolean,
+) => {
+  const filtered: Record<string, V> = {};
```
jaller94, line 28: I think this might protect us from prototype weirdness.

```suggestion
  const filtered: Record<string, V> = Object.create(null);
```

jaller94 (approved):
I'm still worried about potentially messing up Object prototypes here, but I also don't see how it would be exploitable. JavaScript's Object is so weird and unfit for mutating user input.

weeman1337:
> I'm still worried about potentially messing up Object prototypes here, but I also don't see how it would be exploitable. JavaScript's Object is so weird and unfit for mutating user input.

Please bring this up in our discussions. We can talk about enabling the linter and then work on all the places.

packages/react-sdk/src/lib/filterRecord.ts
weeman1337: I will create follow-up tickets to cover all comments about objects with our linting tool.s

PR #773, opened 2025-03-10
Add ADR about text formatting

jaller94:
This document comes from a research spike about how to store formatted text.

### DIFF
--- a/docs/adrs/adr009-text-formatting.md
+++ b/docs/adrs/adr009-text-formatting.md
@@ -1,0 +1,56 @@
+# ADR008: Text formatting
+
+Status: accepted
+
+The text of shapes can be formatted using a restricted set of HTML as a format type.
+
+```json
+{
+  "type": "shape",
+  "kind": "rectangle",
+  "position": {
+    "x": 0,
+    "y": 0
+  },
+  "width": 100,
+  "height": 50,
+  "textFormat": "custom.html",
+  "text": "Hello <strong>World</strong>!"
+}
+```
+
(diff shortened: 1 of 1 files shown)


docs/adrs/adr009-text-formatting.md
```diff
@@ -0,0 +1,56 @@
+# ADR009: Text formatting
+
+Status: accepted
```
mgcm, line 3: For the current scope of research, I would say this should be `draft` instead. Once we implement it, then it can move to `accepted`.

docs/adrs/adr009-text-formatting.md
```diff
@@ -0,0 +1,56 @@
...
+    "y": 0
+  },
+  "width": 100,
+  "height": 50,
+  "textFormat": "custom.html",
+  "text": "Hello <strong>World</strong>!"
+}
+```
+
+This introduces a new property `textFormat` which is either missing or has the value `custom.text`. If it is missing, the property `text` contains plain text. If it has the value `custom.text`, it's expected to be the restricted subset of HTML described in this document.
```
mgcm, line 22: All occurrences of `custom.text` should be changed to `custom.html`, right?

docs/adrs/adr009-text-formatting.md
```diff
@@ -0,0 +1,56 @@
...
+<text>Hello <tspan style="font-size:74.6667px">World</tspan>!</text>
+```
+
+Links can become `<a xlink:href="">`. Ensure this is safe with `rel="noopener"` and `target="_blank"` or whatever is required. I have not looked this up for SVG.
+
+## Considered alternatives
+
+### Using `<b>`, `<em>` etc.
+
+There is no tag which does the reverse, so
```
mgcm, line 52: Is this sentence complete? If yes, I do not understand it. There is no tag which does the reverse of what?

mgcm (changes requested):

PR #779, opened 2025-03-10
Improve zooming and panning with trackpad, mouse and keyboard

jaller94:
This PR improves how we navigate in infinite canvas mode, specifically:

- new zoom max of 400%
- smoother zooming with stepping affected by scale (the higher the zoom, the bigger the zoom step)
- on trackpad, use two fingers to pan and press CTRL to zoom instead
- right click and move mouse to pan (pointer must not be above a shape)
- use mouse wheel to change zoom level
- use CTRL+(plus or equal) / CTRL+minus to change zoom level (or CMD on macOS)
- use CTRL+0 to reset zoom level to 100% (or CMD+0 on macOS)

### DIFF
--- a/packages/react-sdk/src/components/Shortcuts/ZoomShortcuts/ZoomShortcuts.tsx
+++ b/packages/react-sdk/src/components/Shortcuts/ZoomShortcuts/ZoomShortcuts.tsx
@@ -19,6 +19,7 @@ import { useHotkeys } from 'react-hotkeys-hook';
 import { useSvgScaleContext } from '../../Whiteboard/SvgScaleContext';
 import { zoomStep } from '../../Whiteboard/constants';
 import { HOTKEY_SCOPE_WHITEBOARD } from '../../WhiteboardHotkeysProvider';
+import { isMacOS } from '../../common/platform';
 
 export const ZoomShortcuts: React.FC = function () {
   const { updateScale } = useSvgScaleContext();
@@ -32,7 +33,9 @@ export const ZoomShortcuts: React.FC = function () {
   }, [updateScale]);
 
   useHotkeys(
-    ['ctrl+=', 'meta+=', 'ctrl+equal', 'meta+equal', 'ctrl+plus', 'meta+plus'],
+    isMacOS()
+      ? ['meta+equal', 'meta+=', 'meta+plus']
+      : ['ctrl+equal', 'ctrl+=', 'ctrl+plus'],
     handleZoomIn,
     {
       preventDefault: true,
@@ -43,7 +46,7 @@ export const ZoomShortcuts: React.FC = function () {
   );
 
   useHotkeys(
-    ['ctrl+minus', 'meta+minus', 'ctrl+-', 'meta+-'],
+    isMacOS() ? ['meta+minus', 'meta+-'] : ['ctrl+minus', 'ctrl+-'],
     handleZoomOut,
     {
       preventDefault: true,
--- a/packages/react-sdk/src/components/Shortcuts/ZoomShortcuts/ZoomShortcuts.tsx
+++ b/packages/react-sdk/src/components/Shortcuts/ZoomShortcuts/ZoomShortcuts.tsx
@@ -33,27 +33,27 @@ export const ZoomShortcuts: React.FC = function () {
   }, [updateScale]);
 
   useHotkeys(
-    isMacOS()
-      ? ['meta+equal', 'meta+=', 'meta+plus']
-      : ['ctrl+equal', 'ctrl+=', 'ctrl+plus'],
-    handleZoomIn,
-    {
-      preventDefault: true,
-      enableOnContentEditable: true,
-      scopes: HOTKEY_SCOPE_WHITEBOARD,
-    },
-    [handleZoomIn],
-  );
+    '*',
+    (event: KeyboardEvent) => {
+      const modifier = isMacOS() ? event.metaKey : event.ctrlKey;
 
-  useHotkeys(
-    isMacOS() ? ['meta+minus', 'meta+-'] : ['ctrl+minus', 'ctrl+-'],
-    handleZoomOut,
+      if (event.key === '+' && modifier) {
+        event.preventDefault();
+        handleZoomIn();
+      }
+
+      if (event.key === '-' && modifier) {
+        event.preventDefault();
+        handleZoomOut();
+      }
+    },
     {
-      preventDefault: true,
+      preventDefault: false,
       enableOnContentEditable: true,
       scopes: HOTKEY_SCOPE_WHITEBOARD,
+      useKey: true,
     },
-    [handleZoomOut],
+    [handleZoomIn, handleZoomOut],
   );
 
   return null;
--- a/packages/react-sdk/src/components/Shortcuts/ZoomShortcuts/ZoomShortcuts.tsx
+++ b/packages/react-sdk/src/components/Shortcuts/ZoomShortcuts/ZoomShortcuts.tsx
@@ -29,11 +29,13 @@ export const ZoomShortcuts: React.FC = function () {
         (keyboardEvent.ctrlKey || keyboardEvent.metaKey) &&
         keyboardEvent.key === '+'
       ) {
+        keyboardEvent.preventDefault();
         updateScale(zoomStep);
       } else if (
         (keyboardEvent.ctrlKey || keyboardEvent.metaKey) &&
         keyboardEvent.key === '-'
       ) {
+        keyboardEvent.preventDefault();
         updateScale(-zoomStep);
       }
     },
@@ -44,7 +46,7 @@ export const ZoomShortcuts: React.FC = function () {
     '*',
     handleZoom,
     {
-      preventDefault: true,
+      preventDefault: false,
       enableOnContentEditable: true,
       scopes: HOTKEY_SCOPE_WHITEBOARD,
     },
--- a/packages/react-sdk/src/components/Shortcuts/ZoomShortcuts/ZoomShortcuts.tsx
+++ b/packages/react-sdk/src/components/Shortcuts/ZoomShortcuts/ZoomShortcuts.tsx
@@ -16,28 +16,15 @@
 
 import React, { useCallback, useMemo } from 'react';
 import { useHotkeys } from 'react-hotkeys-hook';
-import { useSvgScaleContext } from '../../Whiteboard/SvgScaleContext';
-import { zoomStep } from '../../Whiteboard/constants';
+import { useZoomControls } from '../../../lib';
 import { isMacOS } from '../../common/platform';
 
 export const ZoomShortcuts: React.FC = function () {
-  const { updateScale, setScale } = useSvgScaleContext();
-
-  const handleResetZoom = useCallback(() => {
-    setScale(1.0);
-  }, [setScale]);
-
-  const handleZoomIn = useCallback(() => {
-    updateScale(zoomStep);
-  }, [updateScale]);
-
-  const handleZoomOut = useCallback(() => {
-    updateScale(-zoomStep);
-  }, [updateScale]);
+  const { resetZoom, zoomIn, zoomOut } = useZoomControls();
 
   const macOs = useMemo(() => isMacOS(), []);
 
-  const isEventToApply = useCallback(
+  const hasKeyModifier = useCallback(
     (event: KeyboardEvent) =>
       (macOs && event.metaKey) || (!macOs && event.ctrlKey),
     [macOs],
@@ -46,8 +33,8 @@ export const ZoomShortcuts: React.FC = function () {
   useHotkeys(
     ['+', '='],
     (event: KeyboardEvent) => {
-      if (isEventToApply(event)) {
-        handleZoomIn();
+      if (hasKeyModifier(event)) {
+        zoomIn();
       }
     },
     {
@@ -56,14 +43,14 @@ export const ZoomShortcuts: React.FC = function () {
       useKey: true,
       splitKey: '_', // needed, otherwise '+' is not applied
     },
-    [handleZoomIn, isEventToApply],
+    [zoomIn, hasKeyModifier],
   );
 
   useHotkeys(
     '-',
     (event: KeyboardEvent) => {
-      if (isEventToApply(event)) {
-        handleZoomOut();
+      if (hasKeyModifier(event)) {
+        zoomOut();
       }
     },
     {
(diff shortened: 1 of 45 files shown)


packages/react-sdk/src/components/Whiteboard/SvgCanvas/SvgCanvas.tsx
```diff
@@ -108,6 +109,71 @@ export const SvgCanvas = forwardRef(function SvgCanvas(
...
+  const handleKeyDown = useCallback(
+    (e: KeyboardEvent<SVGSVGElement>) => {
+      const isArrowKey = [
+        'ArrowUp',
+        'ArrowDown',
+        'ArrowLeft',
+        'ArrowRight',
+      ].includes(e.key);
+
+      if (isArrowKey) {
```
maheichyk, line 148: Maybe could move this to some function, a bit repetitive as it happens the same for another method

packages/react-sdk/src/components/Whiteboard/SvgCanvas/SvgCanvas.tsx
```diff
@@ -108,6 +109,71 @@ export const SvgCanvas = forwardRef(function SvgCanvas(
...
+        e.preventDefault();
+        applyKeyboardNavigation();
+      }
+    },
+    [applyKeyboardNavigation],
+  );
+
+  const handleKeyUp = useCallback(
+    (e: KeyboardEvent<SVGSVGElement>) => {
+      const isArrowKey = [
```
maheichyk, line 161: same is arrow key check here

packages/react-sdk/src/components/Whiteboard/SvgCanvas/SvgCanvas.tsx
```diff
@@ -176,9 +242,12 @@ export const SvgCanvas = forwardRef(function SvgCanvas(
                   }),
             }}
             viewBox={`0 0 ${viewportWidth} ${viewportHeight}`}
+            onClick={handleCanvasClick}
             onMouseDown={onMouseDown}
             onMouseMove={handleMouseMove}
-            onWheel={onWheel}
+            onKeyDown={handleKeyDown}
+            onKeyUp={handleKeyUp}
+            tabIndex={0}
```
maheichyk, line 250: I suggest to apply these changed to infinite canvas mode only. 

For both modes it also adds a focus default outline. On non-infinite it can be seen directly. On infinite user needs to scroll to the edge. I think for infinite we should hide the effect.

I also think we should use `tabIndex={-1}` for inf mode. Having `0` adds extra focus to the whole canvas when user switches from top to bottom bars and this focus is not visible because the canvas is infinite.

packages/react-sdk/src/components/Whiteboard/SvgCanvas/SvgCanvas.tsx
```diff
@@ -108,6 +109,71 @@ export const SvgCanvas = forwardRef(function SvgCanvas(
     setContainerDimensions({ width, height });
   }, [width, height, setContainerDimensions]);
 
+  const handleCanvasClick = useCallback(() => {
+    if (svgRef.current) {
+      svgRef.current.focus();
+    }
+  }, []);
```
maheichyk, line 116: This callback seems to be not needed as soon as there is a tab index on svg canvas. Canvas gets focus and accepts keyboard events.

packages/react-sdk/src/components/Whiteboard/SvgCanvas/SvgCanvas.tsx
maheichyk, line 250: would be nice if user could use up and down arrows already when he opens standalone, but not sure if that should be part of this PR

maheichyk (changes requested):
Looks good to me, works fine some smaller changes requested.

HarHarLinks (changes requested):
Testing the widget with infinite canvas `a3b738f8b77bbf8a19e170a55cbc8e93d3064d01` inside of
```
Version von Element: 1.11.98-rc.0
Krypto-Version: Rust SDK 0.10.0 (3cc301d), Vodozemac 0.9.0
```
on Arch Linux with sway 1.10.1 and Firefox 136.0.2 in wayland native mode
- [x] zooming works from ~11% on my setup to 400%, feels generally good.
- using a mouse or ThinkPad X1 Gen? trackpoint
  - [x] right-click dragging works like Miro
  - [x] middle-click dragging works, without accidentally pasting content. This behaves erratically on Miro. I think this is OK, my experience is better on NeoBoard here, but technically we lose access to the standard Linux right-click clipboard here. I'm not sure how much of this behavior is actually my mouse config translating this to scrolling (see below).
  - [ ] scrolling now scrolls instead of zooming. Naively, this is unexpected to me as it does not match Miro where it just zooms, but it does match Excalidraw. Despite the mentioned config this behavior differs from the above.
  - ctrl+scrolling
    - [x] mouse: now zooms instead
    - [ ] trackpoint: browser zoom
- using trackpad of ThinkPad X1 Gen?
  - [x] right-click dragging works like Miro
  - [ ] middle-click dragging does not work at all, instead it pastes my clipboard
  - [ ] 2 finger scrolling works in all 4 directions (instead of zooming). This does not work in Miro, where it just zooms instead. Not sure what to expect, see mouse
  - ctrl+2 finger scrolling
    - [x] vertically works
    - [ ] horizontally zooms the browser instead, really messy experience
  - [ ] ctrl+middle-click scrolling zooms the browser, nothing on the NeoBoard at all
  - [x] 2 finger pinch to zoom works, like Miro
- using keyboard
  - [x] arrow keys scroll the canvas in all 4 directions
  - [x] ctrl+0 resets the zoom to 100% 
  - [ ] ctrl++/+- zooms in/out towards the cursor; zooming towards center might be better UX here since we're not using the mouse. In Miro, ctrl++ triggers browser zoom for me, while ctrl+- zooms away from the center instead of the cursor.

Notably, some behavior diverges from Miro's, according to the PR description by design, which contrasts with the ticket asking for Miro-like behavior. I think the choices made are generally reasonable; if it was my decision, I would find that acceptable.

However some things are not working as intended (e.g. regular scrolling).

my sway mouse config:
```
input type:pointer {
    accel_profile flat
    dwt disabled
    left_handed disabled
    natural_scroll disabled
    pointer_accel -0.4
    scroll_button BTN_MIDDLE
    scroll_method on_button_down
}
```

packages/react-sdk/src/components/Shortcuts/ZoomShortcuts/ZoomShortcuts.tsx
```diff
@@ -0,0 +1,83 @@
...
+  }, [updateScale]);
+
+  useHotkeys(
+    isMacOS() ? ['meta_+', 'meta_='] : ['ctrl_+', 'ctrl_='],
+    () => {
+      handleZoomIn();
+    },
+    {
+      preventDefault: true,
+      enableOnContentEditable: true,
```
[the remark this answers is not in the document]
mgcm: Well, I was expecting this option would allow using the shortcut while editing the text of a shape but apparently not. All other shortcuts are also using this flag. OK with removing it from the zoom shortcuts.

packages/react-sdk/src/components/Shortcuts/ZoomShortcuts/ZoomShortcuts.tsx
```diff
@@ -0,0 +1,83 @@
...
+  }, [updateScale]);
+
+  const handleZoomOut = useCallback(() => {
+    updateScale(-zoomStep);
+  }, [updateScale]);
+
+  useHotkeys(
+    isMacOS() ? ['meta_+', 'meta_='] : ['ctrl_+', 'ctrl_='],
+    () => {
+      handleZoomIn();
```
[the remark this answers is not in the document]
mgcm: We need to check for meta and ctrl within the actual callback, so I have updated it.

mgcm:
@HarHarLinks regarding your feedback, and focusing on what doesn't work for you:

> scrolling now scrolls instead of zooming. Naively, this is unexpected to me as it does not match Miro where it just zooms, but it does match Excalidraw. Despite the mentioned config this behavior differs from the above.

This is a tough one. Using two fingers on the trackpad and using the mouse wheel both trigger the same wheel events on the browser and they are not easily distinguishable.  I could have an easy fix for it which uses the `wheelDeltaY` property from the `WheelEvent` but that is deprecated. Excalidraw and tldraw indeed show the same behaviour as NeoBoard and apparently a "good" solution would require some sort of heuristic on how fast the events come in or in the way their values change step.

> ctrl+scrolling
>    trackpoint: browser zoom

I do not have a trackpoint available - also, our internal ticket does not mention supporting trackpoints. Kicking this into the future.

> middle-click dragging does not work at all, instead it pastes my clipboard

This should be fixed [by this change](https://github.com/nordeck/matrix-neoboard/pull/779/commits/d29d8574ce630dcdf02b48833d6bb9cfef1332ba#diff-5786d409158ecd2cc06cc2fd45b0cb03c65cfc41a5eb4071bdaf78af1073a661R58)

> Two finger scrolling works in all 4 directions (instead of zooming). This does not work in Miro, where it just zooms instead. Not sure what to expect, see mouse

Miro in macOS does 4 direction scrolling when using two fingers. By using a modifier (ctrl or meta) then it zooms. Seems like Miro is broken for you.

> ctrl+2 finger scrolling
>    [ ] horizontally zooms the browser instead, really messy experience

Can't replicate. Pressing CTRL should enable zoom in / zoom out in this use case.

> ctrl++/+- zooms in/out towards the cursor; zooming towards center might be better UX here since we're not using the mouse. In Miro, ctrl++ triggers browser zoom for me, while ctrl+- zooms away from the center instead of the cursor.

Can't replicate. Zoom in / out with the keyboard shortcuts does not zoom towards the cursor for me and the handler code [is not doing that](https://github.com/nordeck/matrix-neoboard/pull/779/files#diff-bf034fd11e2aef8323de63cc25142c91d13a3c29239a1d4610e9b2d5c9f044e9R31) by overriding the default undefined `origin` parameter of [updateScale](https://github.com/nordeck/matrix-neoboard/blob/5aa6fad9f5335e97871b329482694164fe38776e/packages/react-sdk/src/components/Whiteboard/SvgScaleContext/SvgScaleContextProvider.tsx#L146).

For these last three issues that I can't replicate, maybe @MTRNord can take a look?

maheichyk (approved):
Looks already good for me

HarHarLinks:
> This is a tough one.

Yet, zooming simply by mouse wheel might actually be the relevant one here to have working.

> > middle-click dragging does not work at all, instead it pastes my clipboard
> 
> This should be fixed [by this change](https://github.com/nordeck/matrix-neoboard/pull/779/commits/d29d8574ce630dcdf02b48833d6bb9cfef1332ba#diff-5786d409158ecd2cc06cc2fd45b0cb03c65cfc41a5eb4071bdaf78af1073a661R58)

confirmed, great!

> Seems like Miro is broken for you.

correct, sorry if that wasn't clear :)

> > ctrl+2 finger scrolling
> > [ ] horizontally zooms the browser instead, really messy experience
> 
> Can't replicate. Pressing CTRL should enable zoom in / zoom out in this use case.

Just ftr, this behavior has not changed since my last review.

> > ctrl++/+- zooms in/out towards the cursor; zooming towards center might be better UX here since we're not using the mouse. 
> 
> Can't replicate. Zoom in / out with the keyboard shortcuts does not zoom towards the cursor for me and the handler code [is not doing that](https://github.com/nordeck/matrix-neoboard/pull/779/files#diff-bf034fd11e2aef8323de63cc25142c91d13a3c29239a1d4610e9b2d5c9f044e9R31) by overriding the default undefined origin parameter of [updateScale](https://github.com/nordeck/matrix-neoboard/blob/5aa6fad9f5335e97871b329482694164fe38776e/packages/react-sdk/src/components/Whiteboard/SvgScaleContext/SvgScaleContextProvider.tsx#L146).

I may have interpreted the behavior wrong. It seems to zoom towards the canvas center when I would expect it to zoom towards the viewport center. Does that track for you? This might be minor enough to still include.

---

I found this small issue where scrolling by holding the middle mouse button stops and gets stuck if you happen to have the mouse hover over a toolbar such as the elementbar:

https://github.com/user-attachments/assets/3b725985-74ec-42d6-8ef2-068b9b9ed108

This seems like it may be related to the known issue that drawing a line is interrupted when you cross a toolbar.

MTRNord:
on Fedora Linux with kde plasma and Firefox 137.0 (64-Bit) in wayland native mode and https://github.com/MTRNord/qmk_firmware/tree/firmware24/keyboards/zsa/moonlander/keymaps/mtrnord / https://configure.zsa.io/moonlander/layouts/KvGQz/latest/0 (the zsa uploaded one is a tiny little behind but its mostly that my qmk code does auto capitalisation in the keyboard which shouldnt matter here)
- [x] zooming works from 10% on my setup to 400%
- using a mouse
  - [x] right-click dragging works
  - [x] middle-click dragging works but does accidentally paste content. -> Its outside of view and I think it was a click without drag causing it? Or a double click. But I think this is expected and correct? Please confirm @mgcm 
  - [ ] scrolling now scrolls instead of zooming. -> I agree with @HarHarLinks on his comments
  - ctrl+scrolling
    - [x] mouse: now zooms instead
- [x] Ctrl + Plus/Minus/0 works

HarHarLinks:
- [x] zooming relative to viewport
- [x] scrolling stuck on toolbars fix
- scrolling/zooming heuristic works so-so for my specific setup. my trackpad driver(?) will sometimes auto-lock/adjust almost-vertical scrolling to be perfectly vertical, so I sometimes get a surprising mix of both actions. I would say merge this ignoring this issue for now if the behavior is reasonable for @MTRNord.

MTRNord:
Hm it seems the middle click for me broke. It now always moves (as expected) but on middle mouse click up (so stopping to press) it pastes the text. This worked before the recent changes.

MTRNord:
Also my mouse has a tiltable middle mouse which allows to horizontally move. It works mostly fine just that it zooms in 1% first and then goes left or right. I dont think its a huge issue but I noticed it.

packages/react-sdk/src/components/Whiteboard/WhiteboardHost.tsx
```diff
@@ -54,12 +54,14 @@ import { SvgCanvas } from './SvgCanvas';
 import { useWheelZoom } from './SvgCanvas/useWheelZoom';
 
 const WhiteboardHost = ({
+  contentAreaRef,
```
maheichyk, line 57: This ref is passed through several components to achieve the effect to handle wheel events from the `ElementBar` additionally to `SvgCanvas`. So all "zooming" and "panning" wheel events are propagated to the same handler. From code changes it looks quite verbose currently.

From the user perspective the "panning" behavior is ok, but in the case of "zooming" it looks strange a bit. Zooming on `ElementBar` zooms the canvas. Also within current implementation this zooming behavior happens for the other toolbars: Collaboration, Tools... . 

Checking other apps it seems that a reasonable behavior could be to hide `ElementBar` when we do "zooming" and "panning" on `SvgCanvas`. And when a wheel event happens on `ElementBar` we should do prevent default to avoid browser zooming.

maheichyk (changes requested):

HarHarLinks (dismissed):
earlier comment should've been an approval

maheichyk (approved):

PR #797, opened 2025-04-01
Replace the help icon with a proper icon

MTRNord:
New Icon
grafik

Old Icon
grafik

My reasoning is that the old icon is basically an error/fallback icon and it keeps confusing me and its annoying ^^


HarHarLinks:
Before:
image

After:
image

HarHarLinks (dismissed):
lgtm, not confirmed yet by product

maheichyk:
We would need a changeset if approved by product

HarHarLinks:
Product asks: why use the square (HelpCenter) instead of more common round (Help) icon?
image

MTRNord:
> Product asks: why use the square (HelpCenter) instead of more common round (Help) icon? image

Because most of our UI is square. Like the presentation mode button is rounded square, the toolbar is rounded square, the left panel switcher is too.

HarHarLinks:
In Standalone, we use: 
image

MTRNord:
> In Standalone, we use: 
> image
> 

Yeah but compared to the widget it feels like the exception currently. Maybe we have a consistency issue though

HarHarLinks:
Element Web, which we are visually integrating the widget into: 
image

robertgerbauld:
I support the proposal in changing the icon. May I propose however following the current designs implemented by Elment in the settings area? What do you think?
Bildschirmfoto 2025-04-07 um 13 11 10

HarHarLinks:
> I support the proposal in changing the icon. May I propose however following the current designs implemented by Element in the settings area? What do you think? Only the icon I propose, without any text.

I think that NeoBoard only uses solid icons (not outline, like element), therefore I think my earlier proposal is a better fit.

HarHarLinks (dismissed):

maheichyk (approved):

PR #817, opened 2025-05-05
Get preferred foci from .well-known/matrix/client

mgcm:


packages/react-sdk/src/state/communication/discovery/autodiscovery.ts
```diff
@@ -0,0 +1,206 @@
...
+   */
+  public static async getRawClientConfig(
+    baseUrl: string | undefined,
+  ): Promise<IClientWellKnown> {
+    if (!baseUrl || typeof baseUrl !== 'string' || baseUrl.length === 0) {
+      throw new Error("'baseUrl' must be a string of non-zero length");
+    }
+
+    const response = await this.fetchWellKnownObject(
+      `${baseUrl}/.well-known/matrix/client`,
```
HarHarLinks, line 63: take care that [this well-known](https://spec.matrix.org/latest/client-server-api/#well-known-uri) is available on the "server name", e.g. matrix.org, but not necessarily on the actual server domain, e.g. matrix-client.matrix.org.

mgcm: yup, I have adjusted it to use the spec-defined process of extracting the well-known domain from the Matrix UserID.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -145,6 +146,12 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
     await this.removeOwnSession(whiteboardId, sessionId);
   }
 
+  async updateSessionFoci(sessionId: string, foci: RTCFocus[]): Promise<void> {
+    this.refreshOwnSession(sessionId, {
+      foci_preferred: foci,
+    });
+  }
+
```
maheichyk, line 154: We probably can do session refresh inside of `MatrixRtcSessionManagerImpl` if `livekitFoci` update logic would be as part of `MatrixRtcSessionManagerImpl`

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -55,18 +67,32 @@ export class MatrixRtcCommunicationChannel implements CommunicationChannel {
...
+    this.updateLivekitFoci();
+
+    interval(this.fociPollingInterval)
+      .pipe(
+        takeUntil(this.destroySubject),
+        filter(() => document.visibilityState === 'visible'),
+      )
+      .subscribe(() => {
+        this.logger.debug('Polling for updated LiveKit foci');
+        this.updateLivekitFoci();
```
maheichyk, line 93: This update currently happens after user joins the whiteboard, so session is refreshed after connection. I think this could be improved that we first have foci value and then really join. It looks to me that this would easier if livekitFoci is part of `MatrixRtcSessionManagerImpl`.

Would be also nice to have some tests for this behavior and for behavior regarding update of session with new livekitFoci value.

maheichyk (changes requested):
Foci loading works and session updates, I have some suggestions to change, maybe we should address that.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: Yes, it makes sense. I've addressed this in the latest commit.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -55,18 +67,32 @@ export class MatrixRtcCommunicationChannel implements CommunicationChannel {
   };
   private readonly peerConnections: PeerConnection[] = [];
   private sfuConfig: SFUConfig | undefined;
+  private livekitFoci: LivekitFocus[] | undefined;
```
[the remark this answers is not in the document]
mgcm: After moving the session update logic to the manager, I'm not sure why this needs to be observable. We should discuss again the exact use case. I didn't touch ths.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
mgcm: As for the first part of the comment, I do not disagree but I argue that we should do this when address the sync between RTC connection and session state, which honestly to me is not yet 100% clear that it is something we want (what about webrtc fallback or even just no-rtc fallback?).

As for tests, working on it.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -204,13 +230,82 @@ export class MatrixRtcCommunicationChannel implements CommunicationChannel {
...
+   * Otherwise wait for LiveKit foci to be loaded until a timeout.
+   */
+  private async initLiveKitFoci(): Promise<void> {
+    if (this.livekitFoci !== undefined && this.livekitFoci.length > 0) {
+      this.logger.debug('LiveKit foci already known, skipping init');
+      return;
+    }
+
+    this.logger.debug('Waiting for LiveKit foci...');
+    const source = NEVER.pipe(timeout(2500)).pipe(
```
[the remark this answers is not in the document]
mgcm: Yup, now with the `observeFoci` subject, this can change to just getting the first value from that source.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
mgcm: Added tests for foci discovery and session updates.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -257,7 +310,7 @@ export class MatrixRtcCommunicationChannel implements CommunicationChannel {
 
     // Wait for the Widget API and LiveKit servers to be ready
     await this.widgetApiPromise;
-    await this.initLiveKitServer(session);
+    await this.initLiveKitBackend();
```
maheichyk, line 313: I think we should check if the current `handleSessionJoined` needs to be invoked when another user joins. It doesn't look so.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -56,10 +61,12 @@ export class MatrixRtcCommunicationChannel implements CommunicationChannel {
   };
   private readonly peerConnections: PeerConnection[] = [];
   private sfuConfig: SFUConfig | undefined;
+  private activeFocus: LivekitFocus | undefined;
+  private livekitFoci: LivekitFocus[] | undefined;
```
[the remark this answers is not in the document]
mgcm: yup right, thats only what matters to the channel - all the other potential foci are of interest only to the session state.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -68,6 +75,24 @@ export class MatrixRtcCommunicationChannel implements CommunicationChannel {
...
+      .observeFoci()
+      .pipe(takeUntil(this.destroySubject))
+      .subscribe((foci) => {
+        this.livekitFoci = foci;
+        this.logger.debug('LiveKit foci updated', foci[0]);
+        // Check for focus change and restart backend
+        if (
+          this.activeFocus &&
+          foci.length > 0 &&
+          areLiveKitFociEqual(this.activeFocus, foci[0])
```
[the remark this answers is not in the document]
mgcm: after refactor, this code was thrown out - but yes, agreed.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -84,6 +102,7 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
       this.logger.debug('Already joined a whiteboard, must leave first.');
       await this.leave();
     }
+    this.updateLivekitFoci();
```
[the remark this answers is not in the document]
mgcm: new refactored code will only create / refresh session state when we have a valid backend connection.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
```diff
@@ -282,4 +320,105 @@ describe('MatrixRtcSessionManagerImpl', () => {
     await expect(joinedPromise).resolves.toEqual([]);
     await expect(leftPromise).resolves.toEqual([]);
   });
+
+  it('should fetch and emit LiveKit foci on initialization', async () => {
```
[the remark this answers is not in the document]
mgcm: After the refactor, your suggestion is not correct and initializaion is actually the right word here. I will keep it.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
```diff
@@ -282,4 +320,105 @@ describe('MatrixRtcSessionManagerImpl', () => {
...
+            livekit_service_url: 'https://livekit.example.com',
+            livekit_alias: 'room-id',
+          },
+        ],
+      }),
+      expect.any(Object),
+    );
+  });
+
+  it('should update session with new foci when they change', async () => {
```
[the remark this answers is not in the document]
mgcm: agreed, did so where relevant.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
```diff
@@ -282,4 +320,105 @@ describe('MatrixRtcSessionManagerImpl', () => {
...
+            livekit_service_url: 'https://new-livekit.example.com',
+            livekit_alias: 'room-id',
+          },
+        ];
+      },
+    );
+
+    // Manually trigger an update (instead of waiting for the polling interval)
+    // @ts-ignore: access private method for testing
+    rtcSessionManager.updateLivekitFoci();
```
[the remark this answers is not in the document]
mgcm: with the refactor, this code was removed and done the right way by signalling a foci change.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -299,4 +324,42 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
       this.logger.error('Error while sending RTC session', ex);
     }
   }
+
+  private updateLivekitFoci(): void {
```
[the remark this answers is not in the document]
mgcm: not relevant after refactoring.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -45,16 +46,29 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
...
 
   constructor(
     private readonly widgetApiPromise: Promise<WidgetApi> | WidgetApi,
     private readonly sessionTimeout = DEFAULT_RTC_EXPIRE_DURATION,
-  ) {}
+  ) {
+    interval(this.fociPollingInterval)
+      .pipe(
+        takeUntil(this.destroySubject),
+        filter(() => document.visibilityState === 'visible'),
```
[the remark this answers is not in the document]
mgcm: Not happy about duplicating the channel approach here - for now, removed this filter, maybe we can find a cleaner alternative or don't actually care if it's visibile when polling.

packages/react-sdk/src/components/HelpCenterBar/HelpMenu.tsx
```diff
@@ -21,6 +21,7 @@ import { unstable_useId as useId } from '@mui/utils';
 import { MouseEvent, useCallback, useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useGuidedTour } from '../GuidedTour';
+import { embeddedMode } from '../Whiteboard/constants';
```
maheichyk: Not see why we import constant form `Whiteboard` here. `HelpCenterBar` component independent from `Whiteboard` so there is not need to introduce this dependency here.

packages/react-sdk/src/state/communication/connection/types.ts
```diff
@@ -60,3 +60,13 @@ export type PeerConnection = {
...
+
+/**
+ * Extended PeerConnection that can observe connection state changes
+ */
+export type StatefulPeerConnection = PeerConnection & {
+  /**
+   * Returns an observable that emits when the connection state changes.
+   */
+  observeConnectionState(): Observable<string>;
+};
```
maheichyk: Not sure if we need a separate type for that, I would suggest to add this new method to existing `PeerConnection` , but we not have to implement it for `WebRtcPeerConnection`.

packages/react-sdk/src/state/communication/connection/matrixRtcPeerConnection.ts
```diff
@@ -114,15 +117,21 @@ export class MatrixRtcPeerConnection implements PeerConnection {
     return this.connectionId;
   }
 
+  destroy(): void {
+    this.logger.log(`Destroying connection ${this.connectionId}`);
+
+    this.close();
+  }
```
maheichyk: This method is not called

packages/react-sdk/src/state/communication/connection/matrixRtcPeerConnection.ts
```diff
@@ -33,14 +33,19 @@ import {
...
-import { Message, PeerConnection, PeerConnectionStatistics } from './types';
+import {
+  Message,
+  PeerConnectionStatistics,
+  StatefulPeerConnection,
+} from './types';
 import { extractPeerConnectionStatistics } from './utils';
 
-export class MatrixRtcPeerConnection implements PeerConnection {
+export class MatrixRtcPeerConnection implements StatefulPeerConnection {
```
maheichyk: Would be nice to have some tests for `MatrixRtcPeerConnection` we do have tests for  web rtc version.

packages/react-sdk/src/state/communication/connectionStateHandler.ts
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ConnectionState } from 'livekit-client';
+
+type ConnectionStateHandler = () => Promise<void>;
+
+export async function connectionStateHandler(
```
maheichyk: Not sure we need to have it separated this way, instead of just use this code in `MatrixRtcCommunicationChannel`

packages/react-sdk/src/state/communication/discovery/matrixRtcFocus.ts
```diff
@@ -32,3 +36,99 @@ export interface LivekitFocusActive extends RTCFocus {
...
+  } else {
+    logger.debug('No domain provided to fetch .well-known/matrix/client from');
+  }
+
+  return foci;
+}
+
+// As defined in MSC4143
+// https://github.com/matrix-org/matrix-spec-proposals/blob/toger5/matrixRTC/proposals/4143-matrix-rtc.md#choosing-the-value-of-foci_preferred-for-the-mrtcmember-state-event
+export function makeFociPreferred(
```
maheichyk, line 71: Seems to be a pure function, would be nice to have few tests

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -45,16 +52,55 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
   private readonly leaveSubject = new Subject<void>();
   private readonly sessionJoinedSubject = new Subject<Session>();
   private readonly sessionLeftSubject = new Subject<Session>();
-  // This subject is used to track the state of all sessions, allowing us to
-  // detect sessions and manage their lifecycle, including adding expiration dates.
+  private readonly preferredFociSubject = new ReplaySubject<RTCFocus[]>(1);
+  private readonly activeFocusSubject = new ReplaySubject<RTCFocus>(1);
```
maheichyk: Could we use rxjs `defer` here?

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -75,10 +125,39 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
     return this.sessionLeftSubject;
   }
 
+  observePreferredFoci(): Observable<RTCFocus[]> {
+    return this.preferredFociSubject;
+  }
+
+  // TODO: rename to active focus
```
maheichyk: Todo needed?

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -45,16 +52,55 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
...
     private readonly sessionTimeout = DEFAULT_RTC_EXPIRE_DURATION,
-  ) {}
+  ) {
+    interval(this.fociPollingInterval)
+      .pipe(takeUntil(this.destroySubject))
+      .subscribe(async () => {
+        await this.checkForWellKnownFoci();
+      });
+
+    this.observePreferredFoci()
```
maheichyk: Do we need to subscribe here or instead can run the subscription logic when we emit a new value?

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.test.ts
```diff
@@ -210,52 +225,82 @@ describe('MatrixRtcCommunicationChannel', () => {
...
+        localSessionId: 'session-id',
+        peerConnections: [peerConnection],
       });
+
       return Promise.resolve();
     });
 
     // @ts-ignore - Overriding private method
-    channel.initLiveKitServer = mockInitLiveKitServer;
+    channel.initFocusBackend = mockinitFocusBackend;
```
maheichyk: Would be good not to have to work with private methods, there are other places that use ts-ignore for that

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -274,6 +240,16 @@ export class MatrixRtcCommunicationChannel implements CommunicationChannel {
...
       this.peerConnections.push(peerConnection);
 
+      peerConnection.observeConnectionState().subscribe(async (state) => {
+        this.logger.debug('Peer connection state changed:', state);
+
+        await connectionStateHandler(
+          state,
+          this.connect.bind(this),
+          this.disconnect.bind(this),
+        );
```
maheichyk: I wonder if we should just inline the code here, this binding is not that nice, this is done to reuse code with test, I wounder if necessary.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -184,86 +180,56 @@ export class MatrixRtcCommunicationChannel implements CommunicationChannel {
...
+    this.activeFocus = focus;
+    this.sfuConfig = await AutoDiscovery.getSFUConfigWithOpenID(
+      widgetApi,
+      this.activeFocus as LivekitFocus,
+    );
 
-    // only start peer connection for the current session
-    // because matrix rtc is not peer to peer
-    if (this.sfuConfig && session.sessionId === sessionId) {
+    if (this.sfuConfig) {
```
maheichyk: Maybe change to have an early return here? Where we would log the error message before.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -184,86 +180,56 @@ export class MatrixRtcCommunicationChannel implements CommunicationChannel {
...
+    const { userId, deviceId } = widgetApi.widgetParameters;
+    const sessionId = `_${userId}_${deviceId}`;
 
-    this.logger.log('joined', session.sessionId, session.userId);
+    const session = { sessionId: sessionId, userId: userId } as Session;
 
-    if (!sessionId) {
-      this.logger.error('Unknown session id on session join');
-      throw new Error('Unknown session id on session join');
+    if (embeddedMode) {
```
maheichyk: Currently not sure about this embeddedMode, there is also no test for this to be more clear why it is there

maheichyk (changes requested):

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -56,62 +62,40 @@ export class MatrixRtcCommunicationChannel implements CommunicationChannel {
   };
   private readonly peerConnections: PeerConnection[] = [];
   private sfuConfig: SFUConfig | undefined;
+  private activeFocus: RTCFocus | undefined;
```
maheichyk: Do we need `activeFocus` field at all? I see we just read it from `sessionManager.getActiveFocus()`

packages/react-sdk/src/state/communication/connection/matrixRtcPeerConnection.ts
maheichyk: we already have, sorry somehow confused that

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -45,16 +52,55 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
...
+        const newActiveFocus = this.fociPreferred[0];
+        if (this.activeFocus !== newActiveFocus) {
+          this.activeFocus = newActiveFocus;
+          this.activeFocusSubject.next(newActiveFocus);
+        }
+      });
+
+    this.checkForWellKnownFoci().catch((error) => {
+      this.logger.error('Failed to check for well-known foci:', error);
+    });
```
maheichyk, line 73: Suggest to move this up to the place where pooling is scheduled to be explicit

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -45,16 +52,55 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
...
+  private activeFocus: RTCFocus | undefined;
+  private readonly fociPollingInterval = 60 * 1000;
 
   constructor(
     private readonly widgetApiPromise: Promise<WidgetApi> | WidgetApi,
     private readonly sessionTimeout = DEFAULT_RTC_EXPIRE_DURATION,
-  ) {}
+  ) {
+    interval(this.fociPollingInterval)
+      .pipe(takeUntil(this.destroySubject))
```
maheichyk, line 76: We should consider to change here to use `leaveSubject` instead of `destroySubject` and do a pooling similar to how we work with session refresh to avoid unexpected state when there is no room id.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -45,16 +52,55 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
...
+        await this.checkForWellKnownFoci();
+      });
+
+    this.observePreferredFoci()
+      .pipe(takeUntil(this.destroySubject))
+      .subscribe(async (foci) => {
+        this.logger.debug('Preferred foci updated');
+
+        const widgetApi = await this.widgetApiPromise;
+        const roomId = widgetApi.widgetParameters.roomId ?? '';
```
maheichyk: Currently according to the code this can be called when user has no whiteboard opened, so `roomId` can be in undefined state. I suggest to change that that it is only called when board is opened.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -45,16 +52,55 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
...
+
+        // TODO: Implement member focus logic later
+        const memberFocus: LivekitFocus | undefined = undefined;
+
+        this.fociPreferred = makeFociPreferred(
+          memberFocus,
+          foci as LivekitFocus[],
```
maheichyk: I don't understand why we are just type cast here. I think existing typing can be improved that we don't need to or we have to have validators that check we can actually do that.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -45,16 +52,55 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
...
+          roomId,
+        );
+
+        // Check for a new active foci to change to
+        const newActiveFocus = this.fociPreferred[0];
+        if (this.activeFocus !== newActiveFocus) {
```
maheichyk: Why don't we use `!isEqual` here? We should have a test that checks emitting of a new active focus value.

packages/react-sdk/src/state/communication/discovery/matrixRtcFocus.test.ts
```diff
@@ -0,0 +1,272 @@
...
+    });
+  });
+
+  describe('makeFociPreferred', () => {
+    const livekitAlias = '!room-id';
+    const memberFocus: LivekitFocus = {
+      type: 'livekit',
+      livekit_service_url: 'https://member.livekit.example.com',
+      livekit_alias: '!room-id',
+    };
```
maheichyk: `livekit_alias` is not provided to `makeFociPreferred` function when this function is executed by `MatrixRtcSessionManagerImpl`. We should fix the test (and the function argument type) to be able to see what this function actually takes and test this

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -184,116 +184,92 @@ export class MatrixRtcCommunicationChannel implements CommunicationChannel {
...
+    const session = { sessionId: sessionId, userId: userId } as Session;
 
-    if (!sessionId) {
-      this.logger.error('Unknown session id on session join');
-      throw new Error('Unknown session id on session join');
+    if (isEmbedded()) {
+      // Make sure the livekit alias is set to the roomId
+      // This is required because the communication channel may be initialized
+      // with a bogus room id in embedded mode
+      focus.livekit_alias = roomId;
```
maheichyk: We don't need that if `activeFocus` provided by `MatrixRtcSessionManagerImpl` doesn't suffer from this issue. Therefore we should resolve this in `MatrixRtcSessionManagerImpl` and don't bother client code. It should be possible to achieve that inside of `MatrixRtcSessionManagerImpl` by not emitting a new active focus if it is not joined to the room.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -75,10 +125,38 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
...
+    const domain = widgetApi.widgetParameters.userId?.replace(/^.*?:/, '');
+
+    const foci = await getWellKnownFoci(domain);
+    this.logger.debug('Found homeserver foci', JSON.stringify(foci));
+
+    // There was a change in the homeserver's foci
+    if (!isEqual(foci, this.wellKnownFoci)) {
+      this.logger.debug('Homeserver foci changed');
+      this.wellKnownFoci = foci;
+      this.preferredFociSubject.next(foci);
```
maheichyk: The `preferredFociSubject` is used to emit value for `wellKnownFoci` and we also have a field `fociPreferred` that holds another values (not from well known), this make it very confusing.

We should not use abstractions that we don't need to. As written earlier we can get rid of `preferredFociSubject` at all.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -45,16 +52,55 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
...
-  // detect sessions and manage their lifecycle, including adding expiration dates.
+  private readonly preferredFociSubject = new ReplaySubject<RTCFocus[]>(1);
+  private readonly activeFocusSubject = new ReplaySubject<RTCFocus>(1);
   private readonly sessionSubject = new Subject<SessionState>();
   private sessions: StateEvent<RTCSessionEventContent>[] = [];
   private joinState: { whiteboardId: string; sessionId: string } | undefined;
+  private fociPreferred: RTCFocus[] = [];
+  private wellKnownFoci: RTCFocus[] = [];
+  private activeFocus: RTCFocus | undefined;
+  private readonly fociPollingInterval = 60 * 1000;
```
maheichyk: we should have this `fociPollingInterval` as others in constructor, I also wonder if the name is optimal

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -45,16 +52,55 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
...
   private readonly sessionLeftSubject = new Subject<Session>();
-  // This subject is used to track the state of all sessions, allowing us to
-  // detect sessions and manage their lifecycle, including adding expiration dates.
+  private readonly preferredFociSubject = new ReplaySubject<RTCFocus[]>(1);
+  private readonly activeFocusSubject = new ReplaySubject<RTCFocus>(1);
   private readonly sessionSubject = new Subject<SessionState>();
   private sessions: StateEvent<RTCSessionEventContent>[] = [];
   private joinState: { whiteboardId: string; sessionId: string } | undefined;
+  private fociPreferred: RTCFocus[] = [];
+  private wellKnownFoci: RTCFocus[] = [];
```
maheichyk, line 63: We should improve typing to distinguish values stored in the fields, this will also help with code overall.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
```diff
@@ -282,4 +320,105 @@ describe('MatrixRtcSessionManagerImpl', () => {
...
+        {
+          type: 'livekit',
+          livekit_service_url: 'https://livekit.example.com',
+          livekit_alias: 'room-id',
+        },
+      ],
+    ]);
+  });
+
+  it('should update the membership when LiveKit foci are discovered', async () => {
```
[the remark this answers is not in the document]
mgcm: regarding this test, latest description is `should set membership with preferred foci when joining a session`

packages/react-sdk/src/state/communication/discovery/matrixRtcFocus.ts
mgcm: tests have been added in `matrixRtcFocus.test.ts`

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: removed private variable, added to constructor with default value and renamed to `wellKnownPollingInterval`

mgcm: removed private variable, added to constructor with default value and renamed to `wellKnownPollingInterval`

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: you're right - `makeFociPreferred` doesn't care about the focus type and even less about the alias, so I've refactored it to just manipulate RTCFocus types.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: because I suck :) fixed.

packages/react-sdk/src/state/communication/discovery/matrixRtcFocus.test.ts
mgcm: fixed.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: yup, its me confusing rxjs abstractions. I've removed this subject and now a `computeActiveFocus` method should be called before an active focus change might be required.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: As we discussed, I can't find a better way to do this right now. Both of these two do handle the same types and they might differ, so I'm keeping it like this.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: yes, computing the active foci doesn't require the room id, so I've removed this.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: See other commits, this subject has been removed.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
mgcm: removed

packages/react-sdk/src/state/communication/discovery/matrixRtcFocus.test.ts
```diff
@@ -0,0 +1,223 @@
...
+          livekit_service_url: 'https://wellknown2.livekit.example.com',
+        },
+      ]);
+    });
+
+    it('should return member focus first, then well-known foci', () => {
+      const result = makeFociPreferred(memberFocus, wellKnownFoci);
+
+      expect(result).toEqual([
+        memberFocus,
```
maheichyk, line 141: we are explicit in tests about actual value, lets not use `memberFocus` variable and inline it's value, similar you do with `wellKnownFoci` and please update another test 'should add environment variable....'

packages/react-sdk/src/state/communication/discovery/matrixRtcFocus.ts
```diff
@@ -32,3 +37,84 @@ export interface LivekitFocusActive extends RTCFocus {
...
+      livekit_service_url: envFoci,
+    };
+    preferredFoci.push(livekit_config);
+  }
+
+  // Remove duplicates
+  const foci = preferredFoci.filter(
+    (focus, index, self) => index === self.findIndex((f) => isEqual(f, focus)),
+  );
+
```
maheichyk: Maybe use lodash `uniqWith(preferredFoci, isEqual)`?

packages/react-sdk/src/state/communication/discovery/matrixRtcFocus.test.ts
```diff
@@ -0,0 +1,223 @@
...
+        },
+      ];
+
+      const result = makeFociPreferred(memberFocus, duplicateWellKnownFoci);
+
+      expect(result).toEqual([
+        {
+          type: 'livekit',
+          livekit_service_url: 'https://member.livekit.example.com',
+          livekit_alias: livekitAlias,
```
maheichyk: remove `livekit_alias: livekitAlias,`

packages/react-sdk/src/state/communication/discovery/matrixRtcFocus.test.ts
```diff
@@ -0,0 +1,223 @@
...
+        },
+        {
+          type: 'livekit',
+          livekit_service_url: 'https://wellknown1.livekit.example.com',
+          livekit_alias: livekitAlias,
```
maheichyk: remove `livekit_alias: livekitAlias,`

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -75,10 +99,50 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
...
+    if (!isEqual(foci, this.wellKnownFoci)) {
+      this.logger.debug('Homeserver foci changed');
+      this.wellKnownFoci = foci;
+      this.computeActiveFocus();
+    } else {
+      this.logger.debug('No new homeserver foci found');
+    }
+  }
+
+  async computeActiveFocus(): Promise<void> {
```
maheichyk: should be private method

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -184,116 +185,88 @@ export class MatrixRtcCommunicationChannel implements CommunicationChannel {
...
+      throw new Error('Room ID not found in widget parameters');
     }
-  }
 
-  private async handleSessionJoined(session: Session): Promise<void> {
-    const sessionId = this.sessionManager.getSessionId();
+    const activeFocus = {
+      ...this.sessionManager.getActiveFocus(),
+      livekit_alias: widgetApi.widgetParameters.roomId,
+    } as LivekitFocus;
```
maheichyk: We can't typecast `activeFocus` of type `RTCFocus` here to `LivekitFocus`. We have to declare `activeFocus` of type `LivekitFocus` and ensure that what is returned by `activeFocus` is of type `LivekitFocusConfig` (is it?) and then create an instance of `LivekitFocus` providing `roomId` to `livekit_alias`.

We might want to change `sessionManager.getActiveFocus()` to return `LivekitFocusConfig` because this is what we support currently.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -45,16 +51,30 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
   private readonly leaveSubject = new Subject<void>();
   private readonly sessionJoinedSubject = new Subject<Session>();
   private readonly sessionLeftSubject = new Subject<Session>();
-  // This subject is used to track the state of all sessions, allowing us to
-  // detect sessions and manage their lifecycle, including adding expiration dates.
+  private readonly activeFocusSubject = new ReplaySubject<RTCFocus>(1);
   private readonly sessionSubject = new Subject<SessionState>();
   private sessions: StateEvent<RTCSessionEventContent>[] = [];
   private joinState: { whiteboardId: string; sessionId: string } | undefined;
+  private memberFocus: RTCFocus | undefined = undefined;
```
maheichyk, line 61: We don't work yet with `memberFocus` in this PR, I would prefer to pass `undefined` explicitly to `makeFociPreferred` (maybe with TODO) to make sure this field is not really used somewhere else.

maheichyk (changes requested):

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: updated comment to reflect this

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
mgcm: we discussed this today - I think my changes address your concerns.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: agreed - marked as such

packages/react-sdk/src/state/communication/discovery/matrixRtcFocus.test.ts
mgcm: addressed this by adding a mark comment to the class property.

packages/react-sdk/src/state/communication/connectionStateHandler.ts
mgcm: right, after the tests were updated, this no longer makes sense.

packages/react-sdk/src/state/communication/connection/matrixRtcPeerConnection.ts
mgcm: at some point, I adjusted so that it was called but the tests really don't require it.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: not required - reverted this back, we don't need the replay subject after the tests were updated.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.test.ts
mgcm: right, updated.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
mgcm: also not required with the latest test updates.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
```diff
@@ -282,4 +316,126 @@ describe('MatrixRtcSessionManagerImpl', () => {
...
+            livekit_service_url: 'https://livekit.example.com',
+            livekit_alias: 'room-id',
+          },
+        ],
+      }),
+      expect.any(Object),
+    );
+  });
+
+  it('should update membership with new foci when they change', async () => {
```
maheichyk: Doesn't run fine with `it.only`

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
```diff
@@ -282,4 +316,126 @@ describe('MatrixRtcSessionManagerImpl', () => {
     await expect(joinedPromise).resolves.toEqual([]);
     await expect(leftPromise).resolves.toEqual([]);
   });
+
+  it('should set membership with preferred foci when joining a session', async () => {
+    const sessionManager = new MatrixRtcSessionManagerImpl(widgetApi);
```
maheichyk: a bit strange to use `sessionManager` here when `rtcSessionManager` is used everywhere. Seems that the issue is that activeFocus value is emitted already, I suggest to use `rtcSessionManager.getActiveFocus`

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
maheichyk: Overall the test or the title looks a bit confusing to me. The title doesn't state that we would send an updated membership with re-join, but that we do it test. So we don't update membership if we just joined and foci is changed - this is what kind of could be understood from the test title.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
```diff
@@ -282,4 +316,126 @@ describe('MatrixRtcSessionManagerImpl', () => {
...
+
+  it('should not trigger focus update if the active focus is the same', async () => {
+    vi.spyOn(Date, 'now').mockRestore();
+    vi.useFakeTimers();
+
+    const fakeTime = new Date('2023-02-01T10:11:12.345Z');
+
+    vi.setSystemTime(fakeTime);
+
+    await rtcSessionManager.join('whiteboard-id');
```
maheichyk: We don't have to join to get the `activeFocus`.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
```diff
@@ -282,4 +316,126 @@ describe('MatrixRtcSessionManagerImpl', () => {
...
+
+    widgetApi.sendStateEvent.mockClear();
+
+    // Simulate a foci change
+    vi.spyOn(matrixRtcFocus, 'getWellKnownFoci').mockResolvedValue([
+      {
+        type: 'livekit',
+        livekit_service_url: 'https://livekit.example.com',
+      },
+    ]);
```
maheichyk, line 394: Does it really test the use case? `checkForWellKnownFoci` is not get called second time

maheichyk (changes requested):

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
mgcm: good catch. this was due to how the session manager was created before setting up the fake timers and then the interval for polling would not advance. Running the other tests somehow affected this.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
mgcm: right, reworked this.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
mgcm: correct!

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
mgcm: it does now - might not be before due to the fake timers bug.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
mgcm: how about `should re-join and update session with new foci when they change` ?

mgcm: settled on "should re-join and update session with new foci if they change"

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.test.ts
```diff
@@ -210,52 +238,49 @@ describe('MatrixRtcCommunicationChannel', () => {
...
-        Object.values(channel.getStatistics().peerConnections).length,
-      ).toBe(1);
-    });
-
-    expect(channel.getStatistics()).toMatchObject({
-      localSessionId: 'session-id',
-    });
-  });
-
   async function waitForSessionExists() {
```
maheichyk: This method emits and waits for `peerConnection` statistics to be applied, not really sessions. This is confusing, that it has sessions in the name. Could you please rename and write some comment why it is needed.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.test.ts
```diff
@@ -64,87 +94,71 @@ describe('MatrixRtcCommunicationChannel', () => {
...
     createChannel();
   });
 
   afterEach(() => {
     channel.destroy();
-    vi.resetAllMocks();
+    vi.clearAllMocks();
+  });
+
+  it('should create peer connection and statistics on join', async () => {
```
maheichyk:
```suggestion
  it('should join when connected', async () => {
```
similar to 'should leave when disconnected' test already added

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.ts
```diff
@@ -184,116 +184,118 @@ export class MatrixRtcCommunicationChannel implements CommunicationChannel {
...
+      this.logger.error('Unable to retrieve LiveKit SFU configuration');
+      throw new Error('Unable to retrieve LiveKit SFU configuration');
     }
-  }
 
-  private async handleSessionJoined(session: Session): Promise<void> {
-    const sessionId = this.sessionManager.getSessionId();
+    const { userId, deviceId } = widgetApi.widgetParameters;
+    const sessionId = `_${userId}_${deviceId}`;
+    const session = { sessionId: sessionId, userId: userId } as Session;
```
maheichyk:
```suggestion
    const session: Session = { sessionId: sessionId, userId: userId };
```
would also require `userId` to be checked for undefined

maheichyk (changes requested):
Some final comments, otherwise good to me

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.test.ts
mgcm: renamed to "should create peer connection and join when connected" which is what we want.

packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.test.ts
mgcm: right - the communication channel will create a peer connection  and that method should check that it was indeed created, so that we can then mock the connection to the backend and handle connection state changes.

change name to `waitForPeerConnection`

.changeset/wicked-clowns-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+'@nordeck/matrix-neoboard-react-sdk': minor
+---
+
+Get preferred foci for MatrixRTC from homeserver Well-known URI
```
maheichyk: Should we adjust the changeset?

packages/react-sdk/src/state/communication/discovery/matrixRtcFocus.ts
```diff
@@ -32,3 +37,80 @@ export interface LivekitFocusActive extends RTCFocus {
...
+    preferredFoci.push(livekit_config);
+  }
+
+  const foci = uniqWith(preferredFoci, isEqual);
+
+  logger.debug('Final preferred foci:', foci);
+  return foci;
+}
+
+export const isLivekitFocusConfig = (
```
maheichyk, line 113: Could you please add some tests for the function?

packages/react-sdk/src/state/whiteboardInstanceImpl.ts
```diff
@@ -198,7 +199,7 @@ export class WhiteboardInstanceImpl implements WhiteboardInstance {
     if (matrixRtcMode && sessionManager) {
       communicationChannel = new MatrixRtcCommunicationChannel(
         widgetApiPromise,
-        sessionManager,
+        sessionManager as MatrixRtcSessionManagerImpl,
```
maheichyk: Maybe can improve type casting here

maheichyk (changes requested):

.changeset/wicked-clowns-chew.md
mgcm: How about `Discover MatrixRTC foci from /.well-known/matrix/client and sync session state with backend connection status.`

packages/react-sdk/src/state/communication/discovery/matrixRtcFocus.ts
mgcm: Added in the last PR commit.

packages/react-sdk/src/state/whiteboardInstanceImpl.ts
mgcm: yup! updated.

packages/react-sdk/src/state/communication/discovery/types.ts
```diff
@@ -54,3 +55,27 @@ export type SessionManager = {
    */
   destroy(): void;
 };
+
+export type MatrixRtcSessionManager = SessionManager & {
+  /**
+   * Observes sessions in the current whiteboard.
+   */
+  observeSession(): Observable<SessionState>;
```
maheichyk: duplicate `observeSession` is already declared in `SessionManager`

packages/react-sdk/src/state/whiteboardInstanceImpl.ts
```diff
@@ -184,18 +185,22 @@ export class WhiteboardInstanceImpl implements WhiteboardInstance {
     takeUntil(this.destroySubject),
   );
 
-  static create(
+  static create<T extends SessionManager>(
```
maheichyk: Don't see any usefulness to use generics here, sorry was mis-hint from my side. Let's keep it how it is currently.

packages/react-sdk/src/state/communication/discovery/types.ts
mgcm: damned autocomplete :(

packages/react-sdk/src/state/whiteboardInstanceImpl.ts
mgcm: yea, with the type check function, the generic type and the type cast are no longer needed.

maheichyk (approved):
Looks good 🚀

PR #819, opened 2025-05-07
Improve PDF import for Infinite Canvas

mgcm:
Importing a PDF in Infinite Canvas mode now shows all the pages in a grid layout, from left to right, top to bottom. Pages that do not fit the current max canvas height are ignored.

Screenshot 2025-05-07 at 12 41 25

### DIFF
--- a/.changeset/new-rabbits-heal.md
+++ b/.changeset/new-rabbits-heal.md
@@ -1,0 +1,5 @@
+---
+'@nordeck/matrix-neoboard-react-sdk': patch
+---
+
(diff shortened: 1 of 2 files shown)


mgcm:
> Some test for layout calculation would be nice, otherwise good

I didn't invest time in tests on this PR because there are barely any useful tests in the import logic, just some boilerplate stuff for the dialog behaviour. I am wondering what do you think would be useful here, also considering that this layout behaviour will very likely change in the future.

If you have no strong feelings, I would keep it as it is.

PR #820, opened 2025-05-07
Keep the elements connections during copy and paste, duplicate operations

maheichyk:
User can use copy paste shortcuts and duplicate button in the bar. In this case elements connections are preserved.

### DIFF
--- a/packages/react-sdk/src/components/Shortcuts/ClipboardShortcuts/serialization.ts
+++ b/packages/react-sdk/src/components/Shortcuts/ClipboardShortcuts/serialization.ts
@@ -14,8 +14,10 @@
  * limitations under the License.
  */
 
+import Joi from 'joi';
 import { Base64 } from 'js-base64';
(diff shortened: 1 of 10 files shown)


packages/react-sdk/src/components/Shortcuts/ClipboardShortcuts/serialization.ts
```diff
@@ -14,18 +14,32 @@
...
   Record<'text/plain' | 'text/html', string>
 >;
 
 export type WhiteboardClipboardContent = {
-  elements?: Element[];
+  elements?: Elements;
 };
 
+export type IncomingWhiteboardClipboardContent = {
+  elements?: Element[] | Elements;
```
mgcm, line 32: The PR looks good and is working as expected. Only question I have is why we have to support both the Element array and the Elements object? Can't we just drop `Element[]` from this type?

PR #841, opened 2025-05-15
Increase the pdf import resolution's desired size by 2x

MTRNord:

### DIFF
--- a/.changeset/afraid-rules-burn.md
+++ b/.changeset/afraid-rules-burn.md
@@ -1,0 +1,5 @@
+---
+'@nordeck/matrix-neoboard-react-sdk': patch
+---
+
+Increase the output scale to 4 on infinite canvas mode to ensure the images have enough quality on the board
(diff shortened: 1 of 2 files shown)


HarHarLinks (commented):
I feel like this should be done in L133 like the following because iirc that is our variable for "make it a bit higher quality than the viewport tells you".

```
const outputScale: number = infiniteCanvasMode ? 4 : 2;
```

Ideally whatever logic then places the image on the canvas should still use the fixed outputScale = 2; (i.e. "100%")

MTRNord:
> I feel like this should be done in L133 like the following because iirc that is our variable for "make it a bit higher quality than the viewport tells you".
> 
> ```
> const outputScale: number = infiniteCanvasMode ? 4 : 2;
> ```
> 
> Ideally whatever logic then places the image on the canvas should still use the fixed outputScale = 2; (i.e. "100%")

Updated

HarHarLinks (approved):

PR #863, opened 2025-06-10
Use delayed events to terminate Matrix RTC session

maheichyk:


README.md
```diff
@@ -65,6 +65,7 @@ Check the following steps to develop for the widget:
 You need to install Node.js (`>= 20.0.0`, prefer using an LTS version) and run
 `yarn` to work on this package.
 The minimal Element version to use this widget is `1.11.8`.
```
mgcm: I know it's not related to this PR but we should consider updating this to `1.11.46`, which is when support for file upload via Widget API was [added to Element Web](https://github.com/element-hq/element-web/releases/tag/v1.11.46).

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -106,18 +119,50 @@ export class MatrixRtcSessionManagerImpl implements SessionManager {
...
+    }
 
-    await this.refreshOwnSession(sessionId);
+    if (removeSessionDelayId) {
+      interval(this.removeSessionDelay * 0.75)
+        .pipe(
+          takeUntil(this.destroySubject),
+          takeUntil(this.leaveSubject),
+          switchMap(() =>
+            widgetApi.updateDelayedEvent(
```
mgcm, line 228: this mostly fails when removing the widget and then pinning it back using Element Web - it will complain that the delayed event was not found. I've looked for some race condition here but it was inconclusive. We should troubleshoot it further. Was able to make it go away by wrapping this in a try / catch, which doesn't really make sense to me...

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
```diff
@@ -27,6 +29,7 @@ import { MatrixRtcSessionManagerImpl } from './matrixRtcSessionManagerImpl';
 
 describe('MatrixRtcSessionManagerImpl', () => {
   const fixedDate = 1742832000;
+  const removeSessionDelay = 8000;
```
mgcm, line 32: why 8 seconds? given the update timer considers 0.75, it means the hearbeat will be every 6 seconds, which is fine, just wondering how / why you chose this value. Element Call is currently [using 4 seconds](https://github.com/element-hq/element-call/blob/aa092c2195bf95ae02c214531235443f52756ff2/config/config.sample.json#L19) for restarts.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: I guess because catching the error doesn't break the stream, so if there is some server-side issue and the first request can't handle the delay yet, further requests will be processed even if the first one fails.

mgcm (changes requested):

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
maheichyk: I have worked on improving the possible error handling there. If a delayed event fails to refresh (restart) then the id is cancelled. There are some tricky scenarios and I wrote the tests for them.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
maheichyk: No really a strong opinion here. A bit hard to say what gonna be really good values currently. Open to discuss.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
```diff
@@ -210,6 +332,60 @@ describe('MatrixRtcSessionManagerImpl', () => {
...
+
+    expect(widgetApi.updateDelayedEvent).toHaveBeenLastCalledWith(
+      'syd_bcooaGNyKtyFbIGjGMQR',
+      UpdateDelayedEventAction.Restart,
+    );
+    expect(widgetApi.updateDelayedEvent).toHaveBeenCalledTimes(1);
+
+    await vi.advanceTimersByTimeAsync(removeSessionDelay * 0.75);
+
+    expect(widgetApi.updateDelayedEvent).toHaveBeenCalledTimes(1);
```
mgcm, line 386: not sure I understand why this is being tested like this. can you explain it?

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
```diff
@@ -273,6 +439,244 @@ describe('MatrixRtcSessionManagerImpl', () => {
...
+          scope: 'm.room',
+        },
+        { stateKey: '_@user-id_DEVICEID' },
+      ),
+    );
+
+    expect(widgetApi.sendDelayedStateEvent).toHaveBeenCalledTimes(1);
+  });
+
+  it('should update membership and and re-schedule remove membership delayed event when membership is removed by another user', async () => {
```
mgcm: `and and` (repeated) and `reschedule` (without the -)

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
maheichyk: Yes, the `widgetApi.updateDelayedEvent` is mocked in this test to throw an Error. The attempt to restart the delayed event fails. Then in line 384 the timer is advanced again and we check to be explicit that we don't try to refresh a delayed event again, because it will already be cancelled with our timeout config and don't need to try to update it further.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: this has been mitigated with the latest PR updates.  I think we have a fair tradeoff with the current implementation and how it handles the racing associated with refreshing the client and not canceling the existing delayed event, instead restoring the session when it eventually comes ttru.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
mgcm: After testing this, it seems like a reasonable value - this is on average fast enough to quickly convey to other users that someone has left, a few seconds after the fact.

mgcm (dismissed):
👍

PR #870, opened 2025-06-26
Nic/feat/neo 1389 oldest membership

mgcm:


.changeset/mean-shoes-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+'@nordeck/matrix-neoboard-react-sdk': minor
+---
+
+Use active member focus on join and update on session membership changes"
```
maheichyk: I think "oldest membership for active focus" wording from PR title makes more sense as it is more concrete then "active member focus" to me. Could also remove trailing `"`.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -54,10 +54,6 @@ export class MatrixRtcSessionManagerImpl implements MatrixRtcSessionManager {
   private readonly sessionSubject = new Subject<SessionState>();
   private sessions: StateEvent<RTCSessionEventContent>[] = [];
   private joinState: { whiteboardId: string; sessionId: string } | undefined;
-  /*
-   * MARK: memberFocus is currently left undefined until we work on focus updates
-   * resulting from session membership changes
-   **/
   private memberFocus: RTCFocus | undefined = undefined;
```
maheichyk: I wonder if we have to have this `memberFocus` as a field. It looks to me that the way it is used now we can calculate it and don't need to store it in the field.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -198,15 +200,74 @@ export class MatrixRtcSessionManagerImpl implements MatrixRtcSessionManager {
...
     }
   }
 
-  private computeActiveFocus() {
+  private async checkMemberFocus() {
+    const widgetApi = await this.widgetApiPromise;
+    let sessions: StateEvent<RTCSessionEventContent>[] = [];
+
+    try {
+      sessions = await widgetApi.receiveStateEvents(STATE_EVENT_RTC_MEMBER);
```
maheichyk: Why do we load events with `receiveStateEvents` here? Don't we have them in `this.sessions` synced? Is it to eliminate the possibility of race conditions, could you please comment? This all member data could grow quite a lot. This seems to be executed with each timer tick and the session left.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -198,15 +200,74 @@ export class MatrixRtcSessionManagerImpl implements MatrixRtcSessionManager {
...
       this.wellKnownFoci = foci;
-      this.computeActiveFocus();
+      await this.computeActiveFocus();
     } else {
       this.logger.debug('No new homeserver foci found');
     }
   }
 
-  private computeActiveFocus() {
+  private async checkMemberFocus() {
```
maheichyk: I think this method can return `memberFocus` directly, which is the oldest member preferred focus.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
```diff
@@ -402,4 +402,73 @@ describe('MatrixRtcSessionManagerImpl', () => {
       livekit_service_url: 'https://livekit.example.com',
     });
   });
+
+  it('should use member focus when joining ongoing session', async () => {
+    await rtcSessionManager.join('whiteboard-id');
+
+    widgetApi.sendStateEvent.mockClear();
```
maheichyk: mock clear statement seems to be not needed here

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
```diff
@@ -402,4 +402,73 @@ describe('MatrixRtcSessionManagerImpl', () => {
...
+
+    // @ts-ignore forcefully set for tests
+    widgetApi.widgetParameters.deviceId = 'OTHERDEVICEID';
+    const otherSessionManager = new MatrixRtcSessionManagerImpl(widgetApi);
```
maheichyk: I am wonder if we should use another instance of `MatrixRtcSessionManagerImpl` to test this behaviour. In widget we have one instance of it that reacts to member events. Shouldn't we just "send" member event that will already have livekit focus preferred and that one should be used instead of something we have in wellknown?

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
```diff
@@ -402,4 +402,73 @@ describe('MatrixRtcSessionManagerImpl', () => {
...
+
+    expect(
+      matrixRtcFocus.isEqualFocus(otherSessionManager.getActiveFocus(), {
+        type: 'livekit',
+        livekit_service_url: 'https://livekit.example.com',
+      }),
+    ).toEqual(true);
+  });
+
+  it('should change active focus when oldest member leaves', async () => {
```
maheichyk: I think we should add a test that active focus is not changed when not the oldest member leaves.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
```diff
@@ -402,4 +402,73 @@ describe('MatrixRtcSessionManagerImpl', () => {
...
+      {
+        type: 'livekit',
+        livekit_service_url: 'https://other.livekit.example.com',
+      },
+    ]);
+
+    await otherSessionManager.join('whiteboard-id');
+
+    expect(
+      matrixRtcFocus.isEqualFocus(otherSessionManager.getActiveFocus(), {
```
maheichyk: Why we don't explicitly check what is expected to be an active focus? (use `isEqualFocus` instead)

packages/react-sdk/src/state/communication/discovery/matrixRtcFocus.test.ts
```diff
@@ -249,4 +251,99 @@ describe('matrixRtcFocus', () => {
...
+
+      const focus2: RTCFocus = {
+        type: 'livekit',
+        livekit_service_url: 'https://livekit.example.com',
+      };
+
+      expect(isEqualFocus(focus1, focus2)).toBe(true);
+    });
+
+    it('should return true for identical objects with additional properties', () => {
```
maheichyk, line 272: Why do we have to have this specific behaviour? Generally sounds strange to have things equal that are not, maybe I don't follow what we want to achive here. I kind of thiking that we maybe could use base `isEqual` logic from lodash.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: I am doing it this way because the session list only gets populated _after_ you join the session. To join the session, you first need to connect to the backend, so to find out which is target backend we need to query the active sessions first, before joining, at a time where `this.sessions` is always empty because the session state event observer hasn't been setup yet.

mgcm: maybe this can be changed to only use `receiveStateEvents` if the device is not joined yet, then use the `this.sessions` data when members are leaving.

packages/react-sdk/src/state/communication/discovery/matrixRtcFocus.test.ts
mgcm: this is because the member focus is extracted from the `preferred_foci` property of the oldest session state event, and that data has the `livekit_alias´ property, which is not relevant for focus calculations and will fail `isEqual` if we use that.

.changeset/mean-shoes-relax.md
mgcm: new suggestion:

`Implement oldest membership selection type for the active focus when using MatrixRTC with a Livekit backend`

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: sounds like you're right - refactoring that out.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: aligns with removing the `memberFocus` field, yes.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
mgcm: To do so, we would need to remove the single session manager from the `beforeEach()` common test method and send the mocked first / oldest member event before that object is instantied, as it will be checked when `checkForWellKnownFoci` runs. Sounds like a change that affects all the other tests as well.

To me, using a second session manager here reads more like an actual real use case then mocking sending events.

But we can discuss it in sync.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
mgcm: because the active focus is a member focus and has the `livekit_alias` property... lets discuss and find a beter way to address this.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: after discussing this, we decided on the above 👍

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
mgcm: decided on having an empty session manager constructor and doing an explicit initialize that tests can call and `selectWhiteboardManager` also calls.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
mgcm: test added

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -52,31 +52,37 @@ export class MatrixRtcSessionManagerImpl implements MatrixRtcSessionManager {
...
 
-    interval(wellKnownPollingInterval)
+    interval(this.wellKnownPollingInterval)
       .pipe(takeUntil(this.destroySubject))
       .subscribe(async () => {
         await this.checkForWellKnownFoci();
       });
+
+    this.initialized = true;
+  }
```
maheichyk: Looking at this 'initialize' logic again, I don't see why we can't simply have this logic be part of the 'join' method.

We don't need to emit any `activeFocus` until we join as far as I see. In this case we will not need to have changes to `MatrixRtcSessionManager` public API with extra initialize methods.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -198,28 +210,95 @@ export class MatrixRtcSessionManagerImpl implements MatrixRtcSessionManager {
...
     // Check for a new active foci to change to
     const newActiveFocus = this.fociPreferred[0];
     if (!isEqual(this.activeFocus, newActiveFocus)) {
+      this.logger.debug('New active focus:', newActiveFocus);
       this.activeFocus = newActiveFocus;
       this.activeFocusSubject.next(newActiveFocus);
     }
   }
 
+  private async checkMemberFocus(): Promise<RTCFocus | undefined> {
```
maheichyk: The `check` word sounds a bit confusing here, sounds more like validation. Maybe use `find` or `get` as we are finding oldest member and getting its first value from `foci_preferred`?

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
```diff
@@ -198,28 +210,95 @@ export class MatrixRtcSessionManagerImpl implements MatrixRtcSessionManager {
...
+      );
+      try {
+        sessions = await widgetApi.receiveStateEvents(STATE_EVENT_RTC_MEMBER);
+      } catch (error) {
+        this.logger.error(
+          'Failed to receive session member state events',
+          error,
+        );
+        return;
+      }
```
maheichyk, line 346: I can't follow why we have to try to load events via `widgetApi.receiveStateEvents` with this generic error handling.

The `widgetApi.observeStateEvents(STATE_EVENT_RTC_MEMBER)` on join will already implicitly use `receiveStateEvents` to read events and pass them to `handleRTCSessionEvent` to be added to sessions. If we are not in the room, why do we need to run this code at all?

Commenting out this code doesn't break any test.

I think it is an issue similar to initialization that we run when no room is joined.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
```diff
@@ -402,4 +410,146 @@ describe('MatrixRtcSessionManagerImpl', () => {
...
+      state_key: '_@ither-user_OTHERDEVICEID',
+      sender: '@other-user',
+      type: STATE_EVENT_RTC_MEMBER,
+      event_id: '$event-id-0',
+      room_id: '!room-id',
+      origin_server_ts: 0,
+    });
+
+    // Verify the active focus was updated
+    expect(rtcSessionManager.getActiveFocus()).toEqual({
```
maheichyk, line 988: The test passes even if any of events is not sent or timer is not advanced.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: Agreed. What about `selectMemberFocus()`?

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.test.ts
mgcm: I've refactored it so the test is more clear.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: I can't see a way to do this without a significant refactor of the communication channel and how it relies on the session manager to provide the active focus to try to connect to and when the backend focus connection exists, call the `join` method...

My proposal is to use instead the less generic `MatrixRtcSessionManager` type that the `MatrixRtcCommunicationChannel` already is using and add a new `initFociDiscovery()` method that is called in its constructor, replacing the more generic `initialize` approach.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: I've updated a test that showcases why this is needed.

There is a moment where you are not joined yet (no membership state event and no event observers are setup) and you need to check if you are the oldest member or not, to then try use the active focus to connect to the backend, and once that is successful, you are joined and now have the session observers.

maheichyk: Okay, it is clear now, thanks. The only comment now about the code above is that it is invoked when sessions object is empty. 

However, I think it would be better to use `joinState` field to determine this state (`!this.joinState`). Because currently the `selectMemberFocus` is also invoked when a session left is observed and `SessionManager` receives those left events when `join` is called. In this case if `sessions` is empty then it executes `receiveStateEvents` again.

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
maheichyk: Okay, sounds reasonable

packages/react-sdk/src/state/communication/discovery/matrixRtcSessionManagerImpl.ts
mgcm: Ok, makes sense - but to use it, I had to refactor the `joinState` update to be set after the actual state membership event is sent, othewise it's misrepresenting something that hasn't happened yet.

maheichyk (approved):
Looks good 👍

PR #879, opened 2025-07-11
Fix communication channel to stop visibility handling when destroyed

maheichyk:
Stops inner observable from emitting subsequent visibility states if communication channel is destroyed.

### DIFF
--- a/packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.test.ts
+++ b/packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.test.ts
@@ -16,7 +16,6 @@
 
 import { WidgetApi } from '@matrix-widget-toolkit/api';
 import { MockedWidgetApi, mockWidgetApi } from '@matrix-widget-toolkit/testing';
-import { waitFor } from '@testing-library/react';
 import { ConnectionState } from 'livekit-client';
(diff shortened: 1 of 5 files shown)


packages/react-sdk/src/state/communication/matrixRtcCommunicationChannel.test.ts
```diff
@@ -212,19 +250,14 @@ describe('MatrixRtcCommunicationChannel', () => {
...
-  // wait until there is a peer connection in the statistics
-  // this means that the peer connection was created and observables were set up
-  async function waitForPeerConnection() {
-    await waitFor(() => {
-      expect(Object.keys(channel.getStatistics().peerConnections).length).toBe(
-        0,
-      );
-    });
+  // Waits for observers handlers to be registered
+  async function waitForHandlers() {
```
mgcm, line 259: I don't understand how this works :) we should sync on it but not blocking the PR on this.

maheichyk: This is kind of workaround to wait for Peer Connection to be created with the subscription to connection state. Not sure if we should try to improve this in this PR. This works the same as `await Promise.resolve()` used instead of `await waitForHandlers()`

PR #886, opened 2025-08-11
Attach elements to frames

maheichyk:

### DIFF
--- a/docs/model/crdt-documents.md
+++ b/docs/model/crdt-documents.md
@@ -144,6 +144,7 @@ An Element that has a shape attached, that has a text and a fill color.
 | `textSize`       | `number \| undefined`                                                                              | Font size of the text in CSS pixel unit, `undefined` for auto text size.   |
 | `textFontFamily` | `'Inter' \| 'Abel' \| 'Actor' \| 'Adamina' \| 'Chewy' \| 'Gwendolyn' \| 'Pirata One' \| undefined` | The font family of the text. Defaults to `"Inter"`.                        |
 | `connectedPaths` | `string[] \| undefined`                                                                            | The IDs of connected path elements. Currently only lines can be connected. |
+| `attachedFrame`  | `string \| undefined`                                                                              | The ID of the frame this element is attached to.                           |
 
 #### Example
 
@@ -176,6 +177,7 @@ An element that consists of points.
 | `endMarker`             | `'arrow-head-line' \| undefined` | An optional marker for the end of a path.                                                       |
 | `connectedElementStart` | `string \| undefined`            | The ID of connected element on the first point. Currently shapes can be connected.              |
 | `connectedElementEnd`   | `string \| undefined`            | The ID of connected element on the last point. Currently shapes can be connected.               |
+| `attachedFrame`         | `string \| undefined`            | The ID of the frame this element is attached to.                                                |
 
 #### Example
 
@@ -246,15 +248,16 @@ An image element
 
 #### Fields
 
-| Field      | Type                                                            | Description                                                                                                                      |
-| ---------- | --------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- |
-| `type`     | `'image'`                                                       | Identifies the element as an image.                                                                                              |
-| `mxc`      | `string`                                                        | [MXC URI](https://spec.matrix.org/v1.9/client-server-api/#matrix-content-mxc-uris) pointing to the image.                        |
-| `fileName` | `string`                                                        | Image file name.                                                                                                                 |
-| `mimeType` | `'image/gif' \| 'image/jpeg' \| 'image/png' \| 'image/svg+xml'` | Supported image mime type. (Deprecated. We only have this field for backwards compatibility in the model. It isnt used anymore.) |
-| `position` | `Point`                                                         | The position of the image on the whiteboard canvas.                                                                              |
-| `width`    | `number`                                                        | Scaling of the image on the x-axis.                                                                                              |
-| `height`   | `number`                                                        | Scaling of the image on the y-axis.                                                                                              |
+| Field           | Type                                                            | Description                                                                                                                      |
+| --------------- | --------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- |
+| `type`          | `'image'`                                                       | Identifies the element as an image.                                                                                              |
+| `mxc`           | `string`                                                        | [MXC URI](https://spec.matrix.org/v1.9/client-server-api/#matrix-content-mxc-uris) pointing to the image.                        |
+| `fileName`      | `string`                                                        | Image file name.                                                                                                                 |
+| `mimeType`      | `'image/gif' \| 'image/jpeg' \| 'image/png' \| 'image/svg+xml'` | Supported image mime type. (Deprecated. We only have this field for backwards compatibility in the model. It isnt used anymore.) |
+| `position`      | `Point`                                                         | The position of the image on the whiteboard canvas.                                                                              |
+| `width`         | `number`                                                        | Scaling of the image on the x-axis.                                                                                              |
+| `height`        | `number`                                                        | Scaling of the image on the y-axis.                                                                                              |
+| `attachedFrame` | `string \| undefined`                                           | The ID of the frame this element is attached to.                                                                                 |
 
 #### Example
 
@@ -277,12 +280,13 @@ A frame element is used to group elements and run structured presentations.
 
 #### Fields
 
-| Field      | Type      | Description                                           |
-| ---------- | --------- | ----------------------------------------------------- |
-| `type`     | `'frame'` | Identifies the element as a frame.                    |
-| `position` | `Point`   | The position of the element on the whiteboard canvas. |
-| `width`    | `number`  | Scaling of the element on the x-axis.                 |
-| `height`   | `number`  | Scaling of the element on the y-axis.                 |
+| Field              | Type                    | Description                                           |
+| ------------------ | ----------------------- | ----------------------------------------------------- |
+| `type`             | `'frame'`               | Identifies the element as a frame.                    |
+| `position`         | `Point`                 | The position of the element on the whiteboard canvas. |
+| `width`            | `number`                | Scaling of the element on the x-axis.                 |
+| `height`           | `number`                | Scaling of the element on the y-axis.                 |
+| `attachedElements` | `string[] \| undefined` | The IDs of the elements attached to this frame.       |
 
 #### Example
 
@@ -294,6 +298,42 @@ A frame element is used to group elements and run structured presentations.
   "height": 200
 }
 ```
+
+### Attached elements
+
+Elements can be attached to frames via different operations: moving, frame resizing, pasting and
+drag and drop elements to frames. Frames can't be attached to other frames.
+
+Frames apply changes to the attached elements when the frame is is moved or resized or in the case of other operations.
+
+Rectangle shape with id: `E8LVfJCoYubKA-x2cuc0n` that is attached to the frame:
+
+```json
+{
+  "type": "shape",
+  "kind": "rectangle",
+  "position": { "x": 100, "y": 100 },
+  "fillColor": "#ffffff",
+  "height": 200,
+  "width": 200,
+  "text": "Hello World",
+  "attachedFrame": "jmU4s3M4aysDWiSVKAXI2"
+}
+```
+
+Frame element with id `jmU4s3M4aysDWiSVKAXI2` that has shape attached:
+
+```json
+{
+  "type": "frame",
+  "position": { "x": 50, "y": 100 },
+  "width": 100,
+  "height": 200,
+  "attachedElements": ["E8LVfJCoYubKA-x2cuc0n"]
+}
+```
+
+The attached elements and frames have their IDs exported and are used during the import.
 
 ### `Point`
 
--- a/packages/react-sdk/src/components/ElementAttachFrameProvider/ElementAttachFrameProvider.tsx
+++ b/packages/react-sdk/src/components/ElementAttachFrameProvider/ElementAttachFrameProvider.tsx
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2025 Nordeck IT + Consulting GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { createContext, PropsWithChildren, useMemo, useState } from 'react';
+
+export type ElementAttachFrameSetterContextType = {
+  setElementAttachFrame: (elementAttachFrame: Record<string, string>) => void;
+
+  setAttachedElementsMovedByFrame: (
+    attachedElementsMovedByFrame: string[],
+  ) => void;
+
+  setConnectingPathIds: (connectingPathIds: string[]) => void;
+};
+
+export const ElementAttachFrameSetterContext = createContext<
+  ElementAttachFrameSetterContextType | undefined
+>(undefined);
+
+export type ElementAttachFrameGetterContextType = {
+  elementAttachFrame: Record<string, string>;
+  isFrameHasElementMoved: (frameElementId: string) => boolean;
+  isElementMovedHasFrame: (elementId: string) => boolean;
+};
+
+export const ElementAttachFrameGetterContext = createContext<
+  ElementAttachFrameGetterContextType | undefined
+>(undefined);
+
+export function ElementAttachFrameProvider({
+  children,
+}: PropsWithChildren<{}>) {
+  const [elementAttachFrame, setElementAttachFrame] = useState<
+    Record<string, string>
+  >(() => ({}));
+  const [attachedElementsMovedByFrame, setAttachedElementsMovedByFrame] =
+    useState<string[]>(() => []);
+  const [connectingPathIds, setConnectingPathIds] = useState<string[]>(
+    () => [],
+  );
+
+  const context = useMemo<ElementAttachFrameGetterContextType>(
+    () => ({
+      elementAttachFrame,
+      connectingPathIds,
+      isFrameHasElementMoved: (frameElementId) =>
+        Object.entries(elementAttachFrame).some(
+          ([attachmentElementId, attachmentFrameElementId]) =>
+            attachmentFrameElementId === frameElementId &&
+            !attachedElementsMovedByFrame.includes(attachmentElementId) &&
+            !connectingPathIds.includes(attachmentElementId),
+        ),
+      isElementMovedHasFrame: (elementId) =>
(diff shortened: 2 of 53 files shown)


packages/react-sdk/src/components/ElementAttachFrameProvider/ElementAttachFrameProvider.tsx
```diff
@@ -0,0 +1,87 @@
...
+  children,
+}: PropsWithChildren<{}>) {
+  const [elementAttachFrame, setElementAttachFrame] = useState<
+    Record<string, string>
+  >(() => ({}));
+  const [attachedElementsMovedByFrame, setAttachedElementsMovedByFrame] =
+    useState<string[]>(() => []);
+  const [connectingPathIds, setConnectingPathIds] = useState<string[]>(
+    () => [],
+  );
```
jaller94: Why are these functions?

```suggestion
  const [elementAttachFrame, setElementAttachFrame] = useState<
    Record<string, string>
  >({});
  const [attachedElementsMovedByFrame, setAttachedElementsMovedByFrame] =
    useState<string[]>([]);
  const [connectingPathIds, setConnectingPathIds] = useState<string[]>([]);
```

packages/react-sdk/src/components/Shortcuts/ClipboardShortcuts/ClipboardShortcuts.tsx
```diff
@@ -218,15 +206,30 @@ export function ClipboardShortcuts() {
...
+            if (
+              newElement.type !== 'frame' &&
+              newElement.attachedFrame === undefined
+            ) {
+              newElement = copyElementWithAttachedFrame(
+                newElement,
+                slideInstance.getFrameElements(),
+              );
+            }
+            newElements[elementId] = newElement;
```
jaller94, line 229: As far as I know, this is safe and can't lead to [Prototype Polution](https://stackoverflow.com/a/60850027). 👍

Keep coding as if the next 100 boards I'm going to share with you, will contain the `elementId`s `__proto__` and `constructor`. 😈

packages/react-sdk/src/state/crdt/documents/elements.ts
```diff
@@ -343,3 +353,69 @@ export function modifyElementPosition<T extends Element>(
...
+ * @param frameElements - frame elements
+ * @returns frame id
+ */
+export function findFrameToAttach(
+  element: ShapeElement | PathElement | ImageElement,
+  frameElements: Elements<FrameElement>,
+): string | undefined {
+  const { position } = element;
+  const { width, height } =
+    element.type == 'path'
```
jaller94: I recommend turning [eqeqeq](https://eslint.org/docs/latest/rules/eqeqeq) to "error". :)

```suggestion
    element.type === 'path'
```

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/utils.ts
```diff
@@ -271,6 +399,156 @@ function findPathElementsToDisconnect(
...
+      const frameElementId = findFrameToAttach(element, frameElements);
+      if (frameElementId) {
+        elementAttachFrame[elementId] = frameElementId;
+      }
+    }
+  }
+
+  return Object.fromEntries(
+    Object.entries(elementAttachFrame).filter(([elementId, frameId]) => {
+      const element = elements[elementId];
```
jaller94, line 458: TypeScript doesn't cry that `element` could be `undefined`?

Is `"noUncheckedIndexedAccess": true` not set in tsconfig.json?

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/utils.ts
```diff
@@ -271,6 +399,156 @@ function findPathElementsToDisconnect(
...
+        (connectedElementEnd === undefined ||
+          elementAttachFrame[connectedElementEnd] === frameId)
+      );
+    }),
+  );
+}
+
+/**
+ * Find elements that were moved because the attached frame moved.
+ * @param elements elements
```
jaller94:
```suggestion
```

packages/react-sdk/src/state/whiteboardSlideInstanceImpl.ts
```diff
@@ -305,6 +407,58 @@ export class WhiteboardSlideInstanceImpl implements WhiteboardSlideInstance {
...
+              attachElementIds: [newElementId],
+              detachElementIds: [],
+            });
+            if (patch) {
+              updates.push({
+                elementId: attachedFrame,
+                patch: patch,
+              });
+            }
+          } else if (elements[attachedFrame]?.type === 'frame') {
```
jaller94, line 452: I think that this will be something incorrect or unexpected, if `attachedFrame` is "__proto__" or "constructor".
As far as I can see, `attachedFrame` is user input.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/utils.ts
```diff
@@ -271,6 +399,156 @@ function findPathElementsToDisconnect(
...
+      changeFrameElements[newFrameId] = {
+        attachElementIds: [],
+        detachElementIds: [],
+      };
+    }
+    if (oldFrameId) {
+      changeFrameElements[oldFrameId].detachElementIds.push(elementId);
+    }
+    if (newFrameId) {
+      changeFrameElements[newFrameId].attachElementIds.push(elementId);
```
jaller94: This is only safe, if we are 100% sure that these keys have been set by us and are not the default values of an Object created with a prototype and a constructor. Seems to be the case in the lines above.. but it's getting pretty complicated here.

Needless to say, my `frameId`s also include `__proto__` and `constructor`. 😈

packages/react-sdk/src/state/crdt/documents/elements.ts
maheichyk: We should do that at some point. Agree.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/utils.ts
maheichyk: It is not set in tsconfig.json. Generally I agree that this makes sense, especially for this simple use case. 

However I will not do this config change in this PR as there is other code that needs to be changed because of that and this PR is already not small.

maheichyk: Extra `undefined` check is added, though `undefined` could not happen in this place.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/utils.ts
maheichyk: It is updated

packages/react-sdk/src/state/whiteboardSlideInstanceImpl.ts
maheichyk: Validation is improved to check for those undesired values. It now happens during import, copy paste, document snapshots, document changes via rtc. 

We have other similar code changes in the project. So should be addressed as well.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/utils.ts
maheichyk: Added the changes to validate element ids coming from all the sources.

packages/react-sdk/src/components/Shortcuts/ClipboardShortcuts/ClipboardShortcuts.tsx
maheichyk: Validation is improved to check for those undesired values. It now happens during import, copy paste, document snapshots, document changes via rtc.

.changeset/fast-crabs-work.md
```diff
@@ -0,0 +1,6 @@
+---
+'@nordeck/matrix-neoboard-react-sdk': minor
+'@nordeck/matrix-neoboard-widget': minor
+---
+
+Attach elements to the frame. Implement operations on the attached elements. Works when the `REACT_APP_INFINITE_CANVAS` feature flag is enabled.
```
mgcm:
```suggestion
Add attachment / detachment of elements to frames. Implement resizing and moving of frames and how it affects attached / detached elements.
```

docs/model/crdt-documents.md
```diff
@@ -295,6 +299,41 @@ A frame element is used to group elements and run structured presentations.
 }
 ```
 
+### Attached elements
+
+Elements can be attached to frames through various operations, such as moving, resizing the frame, pasting and dragging and dropping elements onto frames. Frames cannot be attached to other frames.
+
+Frames apply changes to their attached elements when a frame is moved or resized or in the case of other operations.
+
+Rectangle shape with id: `E8LVfJCoYubKA-x2cuc0n` that is attached to the frame:
```
mgcm:
```suggestion
Rectangle shape with id: `E8LVfJCoYubKA-x2cuc0n` that is attached to a frame:
```

packages/react-sdk/src/components/ElementAttachFrameProvider/ElementAttachFrameProvider.tsx
```diff
@@ -0,0 +1,85 @@
...
+  setConnectingPathIds: (connectingPathIds: string[]) => void;
+};
+
+export const ElementAttachFrameSetterContext = createContext<
+  ElementAttachFrameSetterContextType | undefined
+>(undefined);
+
+export type ElementAttachFrameGetterContextType = {
+  elementAttachFrame: Record<string, string>;
+  isFrameHasElementMoved: (frameElementId: string) => boolean;
```
mgcm, line 57: these are very confusingly named. lets discuss if these can be made clearer.

packages/react-sdk/src/components/Shortcuts/ClipboardShortcuts/serialization.ts
```diff
@@ -33,10 +38,10 @@ export type IncomingWhiteboardClipboardContent = {
 };
 
 const elementsSchema = Joi.object()
-  .pattern(Joi.string(), elementSchema)
+  .pattern(Joi.string().not(...disallowElementIds), elementSchema)
```
mgcm, line 41: why was this necessary?

packages/react-sdk/src/components/Shortcuts/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+  WhiteboardSlideInstance,
+  findActiveAndAttachedElementIds,
+} from '../../state';
+
+/**
+ * Select active and attached element for each active element in the document order.
+ * @param slideInstance slide instance
+ * @returns active and attached elements of each active element
+ */
+export function selectActiveAndAttachedElementsInDocumentOrder(
```
mgcm: this is a bit long  - how about `selectActiveAndAttachedSortedElements`? and why `select` ? could be `getActiveAndAttachedSortedElements`...

packages/react-sdk/src/state/utils.test.ts
```diff
@@ -160,6 +170,261 @@ describe('findConnectingPaths', () => {
   });
 });
 
+describe('findActiveAndAttachedElementIds', () => {
+  it('should find active together with attached elements', () => {
```
mgcm:
```suggestion
  it('should find both active and attached elements', () => {
```

packages/react-sdk/src/state/utils.ts
```diff
@@ -62,6 +68,158 @@ export function findConnectingShapes(elements: Elements): string[] {
   return uniq(elementIds);
 }
 
+/**
+ * Return the ids of the active elements and the ids of the attached elements for each active element.
+ * @param activeElementIds - active elements
+ * @param frameElements - frame elements
+ * @returns active and attached elements of each active element
+ */
+export function findActiveAndAttachedElementIds(
```
mgcm, line 77: I find this a bit more concise:

```
export function findActiveAndAttachedElementIds(
  activeElementIds: string[],
  frameElements: Elements<FrameElement>,
): string[] {
  return activeElementIds.flatMap((elementId) => {
    const attached: string[] | undefined =
      frameElements[elementId]?.attachedElements;

    return attached?.length
      ? [elementId, ...attached.filter((id) => !activeElementIds.includes(id))]
      : elementId;
  });
}
```
what do you think?

packages/react-sdk/src/state/utils.ts
```diff
@@ -62,6 +68,158 @@ export function findConnectingShapes(elements: Elements): string[] {
...
+         */
+        elementDetachFrame[elementId] = frameId;
+      }
+    }
+  }
+
+  return elementDetachFrame;
+}
+
+export function invertElementAttachFrame(
```
mgcm: I understand what this is doing but readibility is an issue for me. Just a quick 'n' dirty type alias of FrameID and ElementID to string would make this method signature more obvious:

```
export function getFrameAttachments(
    attachedElements: Record<ElementID, FrameID>,
): Record<FrameID, ElementID[]> {

...

```

packages/react-sdk/src/state/crdt/documents/elements.ts
```diff
@@ -91,7 +96,8 @@ const shapeElementSchema = elementBaseSchema
     stickyNote: Joi.boolean(),
     textSize: Joi.number().strict(),
     textFontFamily: Joi.string().strict().default('Inter'),
-    connectedPaths: Joi.array().items(Joi.string()),
+    connectedPaths: Joi.array().items(Joi.string().not(...disallowElementIds)),
+    attachedFrame: Joi.string().not(...disallowElementIds),
```
mgcm, line 100: we should sync on why the disallowed element ids are also forbidden for strings. I think I get it why it should be for arrays / dicts but not for simple types...

packages/react-sdk/src/components/ElementAttachFrameProvider/ElementAttachFrameProvider.tsx
mgcm: We agree to have comments added to this type so its made more clear what they are for.

packages/react-sdk/src/components/Shortcuts/utils.ts
mgcm: we propose changing to `selectActiveAndAttachedOrderedElements`

packages/react-sdk/src/components/Shortcuts/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+import {
+  Elements,
+  WhiteboardSlideInstance,
+  findActiveAndAttachedElementIds,
+} from '../../state';
+
+/**
+ * Select active and attached element for each active element in the document order.
```
mgcm: let's try to rephrase this

packages/react-sdk/src/components/elements/frame/Display.tsx
```diff
@@ -41,7 +42,11 @@ export const Display: React.FC<DisplayProps> = ({
         y={frameProps.position.y}
         fill={theme.palette.common.white}
         height={frameProps.height}
-        stroke={theme.palette.divider}
+        stroke={
+          frameHasElementMoved
+            ? theme.palette.primary.main
+            : theme.palette.divider
```
mgcm:
```suggestion
            : theme.palette.grey[500]
```

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Moveable/MoveableElement.tsx
```diff
@@ -184,15 +234,21 @@ export function MoveableElement({
       const newElementOverrideUpdates = isShowGrid
         ? snapToGridElementOverrideUpdates(
             elementOverrideUpdates,
-            overrides,
+            elements,
             connectingPathElements,
           )
         : elementOverrideUpdates;
 
+      const elementFrameChanges = findElementFrameChanges(elementAttachFrame, {
```
mgcm, line 242: the current logic is attaching an element even if it is behind a frame, which then makes it essentially disappear.

I would expect that if you drag an element that is behind a frame, into a frame, then its over and in front of the frame, and this should be done in a separate ticket.

for now, I suggest that we don't attach an element if its behind the frame.

mgcm: the alternative would be to have a transparent frame and still have the element attached.

mgcm: or also we can enable the order menu for frames (so bring to back can be applied on frames)

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Moveable/MoveableElement.tsx
```diff
@@ -184,15 +234,21 @@ export function MoveableElement({
...
+        ...elements,
+        ...connectingPathElements,
+      });
+
       const updates = elementsUpdates(
```
mgcm, line 247: found a bug related to this logic:

when having a smaller frame within a larger frame with an object attached- when I resize the smaller frame into the object, the object is not attached to the smaller frame (expected, because its already attached to the larger one) but as soon as I move the larger frame, the object becomes attached to the smaller frame.

https://github.com/user-attachments/assets/b9d3da8c-404f-4444-b721-ba7ee38d5938

packages/react-sdk/src/components/ImageUpload/addImagesToSlide.ts
```diff
@@ -105,5 +106,12 @@ export function addImagesToSlide(
       };
     });
   }
-  slide.addElements(images);
+  images = images.map((image) =>
+    copyElementWithAttachedFrame(image, slide.getFrameElements()),
```
mgcm, line 110: another bug around drag / dropping images to a frame: only the last dropped image will be attached.

maheichyk: It is fine now. It was also happening when pasting several elements to the frame.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Moveable/MoveableElement.tsx
maheichyk: It is fixed, we basically now check that element is attached to the frame and is not re-attached in this case. However is a bit tricky because we have re-attaching to work when an element is moved to another frame or when resize of frame happens.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Moveable/MoveableElement.tsx
maheichyk: I think it is a good idea to enable the menu, enabled.

packages/react-sdk/src/components/ElementAttachFrameProvider/ElementAttachFrameProvider.tsx
maheichyk: Updated jsdoc for context types, should be easier to understand, please have a look

packages/react-sdk/src/state/utils.ts
maheichyk: looks good, applied

packages/react-sdk/src/state/utils.ts
maheichyk: This function is renamed to what you have suggested. Also another one similar is renamed too to follow the same pattern. 

I haven't changed anything regarding types alias as I don't see too much value as they both are ids that are strings. 

I have updated jsdoc. I also think this is not that confusing because for element id we have an array in the type.

mgcm (approved):
🚀

PR #888, opened 2025-08-13
Improve touchpad and mouse wheel handing

mgcm:
This PR improves touchpad and mouse wheel handing.

This is now done using system-dependend heuristics.

Because you can't tell if the events are triggered by the mouse wheel or a touchpad and the event deltas very a lot between browsers and operating systems, this implementation supports:

- zoom in / out when using a mouse wheel
- two finger swipe gestures in the touchpad do up/down/left/right panning
- up and down two finger swipe gestures in the touchpad with a modifier key pressed (control or meta) do zoom in / out

where supported (macOS and Linux+Firefox):

- two finger pinching gestures in the touchpad do zoom in / out

**NOTE**
 - this also fixes a bug where middle mouse button / wheel clicking would paste the clipboard contents in the whiteboard

### DIFF
--- a/packages/react-sdk/src/components/Whiteboard/SvgCanvas/SvgCanvas.tsx
+++ b/packages/react-sdk/src/components/Whiteboard/SvgCanvas/SvgCanvas.tsx
@@ -191,6 +191,16 @@ export const SvgCanvas = function ({
     [applyKeyboardNavigation, enableShortcuts],
   );
 
+  // Handle auxiliary mouse button clicks (middle/right click)
+  // To prevent default paste clipboard action on firefox
+  const handleMouseAuxClick: MouseEventHandler<SVGSVGElement> = useCallback(
+    (e) => {
+      e.nativeEvent.preventDefault();
+      e.nativeEvent.stopPropagation();
+    },
+    [],
+  );
+
   const handleMouseMove: MouseEventHandler<SVGSVGElement> = useCallback(
     (e) => {
       const position = calculateSvgCoordsFunc({
@@ -254,6 +264,7 @@ export const SvgCanvas = function ({
           onKeyUp={infiniteCanvasMode ? handleKeyUp : undefined}
           tabIndex={infiniteCanvasMode ? -1 : undefined}
           onMouseLeave={onMouseLeave}
+          onAuxClick={handleMouseAuxClick}
           transform-origin={
             infiniteCanvasMode && !preview ? 'center center' : undefined
           }
--- a/packages/react-sdk/src/components/Whiteboard/SvgCanvas/SvgCanvas.tsx
+++ b/packages/react-sdk/src/components/Whiteboard/SvgCanvas/SvgCanvas.tsx
@@ -195,8 +195,8 @@ export const SvgCanvas = function ({
   // To prevent default paste clipboard action on firefox
   const handleMouseAuxClick: MouseEventHandler<SVGSVGElement> = useCallback(
     (e) => {
-      e.nativeEvent.preventDefault();
-      e.nativeEvent.stopPropagation();
+      e.preventDefault();
+      e.stopPropagation();
     },
     [],
   );
--- a/packages/react-sdk/src/components/Whiteboard/SvgCanvas/useWheelZoom.ts
+++ b/packages/react-sdk/src/components/Whiteboard/SvgCanvas/useWheelZoom.ts
@@ -35,6 +35,9 @@ type UseWheelZoomResult = {
 
 type Timeout = ReturnType<typeof setTimeout>;
 
+/**
+ * This method handles touchpad and scroll wheel mouse events only.
+ */
 export const useWheelZoom = (
   svgRef: RefObject<SVGSVGElement>,
 ): UseWheelZoomResult => {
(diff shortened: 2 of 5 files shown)


packages/react-sdk/src/components/Whiteboard/SvgCanvas/useWheelZoom.ts
```diff
@@ -127,6 +131,64 @@ export const useWheelZoom = (
     [svgRef, scale, updateScale, updateTranslation, resetInteractionMode],
   );
 
+  const handleLinuxWheelZoom = useCallback(
```
maheichyk: Why do we have both `handleMacOSWheelZoom` and `handleMacOSWheelZoom` separate callbacks? 

One cached callback is not be used depending on the OS version. Also logic about wheel zoom in progress is duplicated because of that.

packages/react-sdk/src/components/Whiteboard/SvgCanvas/useWheelZoom.ts
```diff
@@ -127,6 +131,64 @@ export const useWheelZoom = (
...
+      if (wheelZoomTimeoutRef.current) {
+        clearTimeout(wheelZoomTimeoutRef.current);
+        wheelZoomTimeoutRef.current = undefined;
+      }
+      wheelZoomTimeoutRef.current = setTimeout(() => {
+        setWheelZoomInProgress(false);
+      }, 300);
+
+      // Heuristic for zoom vs pan
+      const isZoom =
```
maheichyk: Would be good to have tests for that. This logic is not always possible to debug for everyone as we have different os, hardware.

Would be really nice to have the tests also for existing zooming/touching code already as it is the same functionality from user perspective.

packages/react-sdk/src/components/Whiteboard/SvgCanvas/useWheelZoom.ts
```diff
@@ -139,8 +201,10 @@ export const useWheelZoom = (
     };
   }, []);
 
+  const systemHandler = isMacOS() ? handleMacOSWheelZoom : handleLinuxWheelZoom;
```
maheichyk: Can memo `isMacOS` similar we do in `ZoomShortcuts`. A small 'improvement', but why would we need to calculate it each time?

packages/react-sdk/src/components/Whiteboard/SvgCanvas/useWheelZoom.ts
mgcm: thats right. i've refactored this.

packages/react-sdk/src/components/Whiteboard/SvgCanvas/useWheelZoom.ts
mgcm: I've added a bunch of tests to the `useZoomWheel` scope, maybe its good enough for now?

packages/react-sdk/src/components/Whiteboard/SvgCanvas/SvgCanvas.tsx
```diff
@@ -191,6 +191,16 @@ export const SvgCanvas = function ({
     [applyKeyboardNavigation, enableShortcuts],
   );
 
+  // Handle auxiliary mouse button clicks (middle/right click)
+  // To prevent default paste clipboard action on firefox
+  const handleMouseAuxClick: MouseEventHandler<SVGSVGElement> = useCallback(
+    (e) => {
+      e.nativeEvent.preventDefault();
+      e.nativeEvent.stopPropagation();
```
maheichyk: Why not to use `preventDefault` and `stopPropagation` directly without `nativeEvent`?

packages/react-sdk/src/components/Whiteboard/SvgCanvas/useWheelZoom.ts
```diff
@@ -32,41 +34,82 @@ type UseWheelZoomResult = {
...
-      if (wheelZoomTimeoutRef.current) {
-        clearTimeout(wheelZoomTimeoutRef.current);
-        wheelZoomTimeoutRef.current = undefined;
-      }
-      wheelZoomTimeoutRef.current = setTimeout(() => {
-        setWheelZoomInProgress(false);
-      }, 300);
+      const zoomOriginOnCanvas = calculateSvgCoords(
+        { x: event.clientX, y: event.clientY },
+        svgRef.current!,
```
maheichyk: Suggest not to use the non-null assertion operator. Instead, check that the value is defined when the method is invoked.

packages/react-sdk/src/components/Whiteboard/SvgCanvas/SvgCanvas.tsx
mgcm: good point, I must have followed the logic used a bit below on `handleMouseMove`... removing that.

packages/react-sdk/src/components/Whiteboard/SvgCanvas/useWheelZoom.ts
mgcm: right, updating!

PR #892, opened 2025-08-20
Update widget port to run locally

maheichyk:
Follow up to this discussion: https://github.com/nordeck/matrix-poll/pull/683#discussion_r2219054555

### DIFF
--- a/.changeset/fifty-parts-study.md
+++ b/.changeset/fifty-parts-study.md
@@ -1,0 +1,5 @@
+---
+'@nordeck/matrix-neoboard-widget': patch
+---
+
+Update widget port to run locally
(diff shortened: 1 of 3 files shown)


HarHarLinks (commented):
I dislike this idea:
When I have a local dev env set up, and I start the widget `yarn dev`, then it will be available at `http://localhost:5273/`. I can `/addwidget http://localhost:5273/` and use it in Element Web. Now I start the widget wit `yarn dev:https`, and it will be available at `https://localhost:5273/`. My room that already has the widget set up breaks with a http/https mismatch issue, but finding that issue is obfuscated. I check for using the right port etc, it is all correct.

Since I have to `/addwidget https://something:abcd` anew anyway when running in https mode, I think keeping the current state is preferable - I don't understand the benefit of using the same port for both.

maheichyk:
We have discussed this in another PR and this change is already applied for the poll widget. 

I am fine to have different ports for http/https. Just want to have the same config for all the widgets.

PR #912, opened 2025-09-23
fixes panning pause when pointer is over a shape

mgcm:
This PR fixes a bug where panning the whiteboard with a right-click would pause if the action started outside of a shape and the pointer then moved over a shape.

This is because the panning logic is on this lower layer component (`UnSelectecElementHandler`) and gets overriden by other mouse event listeners on top-most components. We have an internal ticket to refactor this and remove the panning from here but this fixes it for now.

### DIFF
--- a/packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Selection/UnSelectElementHandler.tsx
+++ b/packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Selection/UnSelectElementHandler.tsx
@@ -78,26 +78,12 @@ export function UnSelectElementHandler() {
     ],
   );
 
-  const handleMouseUp = useCallback((event: MouseEvent<SVGRectElement>) => {
-    if (event.button === 1 || event.button === 2) {
-      // Middle and Right click
-      event.preventDefault();
-      event.stopPropagation();
-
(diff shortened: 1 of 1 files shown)


packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Selection/UnSelectElementHandler.tsx
```diff
@@ -128,12 +112,35 @@ export function UnSelectElementHandler() {
...
+
+      updateTranslation(
+        event.clientX - previousPanCoordinates.x,
+        event.clientY - previousPanCoordinates.y,
+      );
+
+      setPreviousPanCoordinates({ x: event.clientX, y: event.clientY });
+    };
+
+    window.addEventListener('mousemove', handleMouseMove);
```
maheichyk, line 133: We should also consider that the mouse can be up when a cursor is over a shape. Currently, in this case, `panEnabled` will not be set to false, and the board will remain in a panning state when the mouse moves.

mgcm: right, I've now also moved the mouse up to the window listener.

PR #924, opened 2026-01-15
Add block arrow shape

nurjinjafar:
This pull request adds a new shape (Block arrow) to the toolbar.

Bildschirmfoto 2026-02-10 um 00 01 36


mgcm (changes requested):
The implementation matches the ticket requirements but I think a few more tests should be modified to also include the block arrow shape:

[ToolsBar.test.tsx](https://github.com/nordeck/matrix-neoboard/blob/main/packages/react-sdk/src/components/ToolsBar/ToolsBar.test.tsx) -> `should render without exploding` and `should disable all buttons if the slide is locked and not active` do not include the new block arrow button

[elements.test.ts](https://github.com/nordeck/matrix-neoboard/blob/main/packages/react-sdk/src/state/crdt/documents/elements.test.ts) -> `should accept %j shape event` and  `should accept %j shape event with attached frame` do not include the block arrow shape type

I know this is not part of the ticket but as an arrow-type tool, I was expecting two different behaviors:

1. the fill color should default to the same as other arrow / line tools, instead of the other shapes' color
2. the left/right arrow tip options are also available for this shape

but maybe this has been a conscious product decision I am not aware of.

Finally, there is a bug in the PDF export, as the shape text is not being limited to the inner part of the block arrow:

 - on the board canvas:
Image

- exported to PDF:
Image

beyond this, looking great!

.changeset/two-parents-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@nordeck/matrix-neoboard-react-sdk': minor
+---
+
+add new Block Arrow shape
```
maheichyk: Add block arrow shape

packages/react-sdk/src/components/BoardBar/pdf/createWhiteboardPdfElementShape.ts
```diff
@@ -106,3 +110,42 @@ function createElementShapeTriangle(element: ShapeElement): Content {
...
+
+  const { position, width, height } = element;
+
+  const arrowHeadWidth = Math.max(width * 0.35, 10);
+  const bodyWidth = width - arrowHeadWidth;
+
+  const inset = height * 0.25;
+  const bodyTop = position.y + inset;
+  const bodyBottom = position.y + height - inset;
+  const centerY = position.y + height / 2;
```
maheichyk: This looks like copied from Display, need some test for the code.

packages/react-sdk/src/components/elements/block-arrow/Display.tsx
```diff
@@ -0,0 +1,153 @@
...
+  'data-testid': dataTestid,
+  ...shape
+}: BlockArrowElementProps) => {
+  const { strokeColor, strokeWidth, text } = getRenderProperties(shape);
+
+  const { x, y } = shape.position;
+  const width = shape.width;
+  const height = shape.height;
+  // Arrowhead width (proportional)
+  const arrowHeadWidth = Math.max(width * 0.35, 10);
```
maheichyk: Looks strange, why you do `max` here. What if the width is small? Then arrow head width will get a fixed width and go outside of of the shape.

packages/react-sdk/src/components/elements/block-arrow/Display.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const bodyWidth = width - arrowHeadWidth;
+
+  const inset = height * 0.25;
```
maheichyk: A bit confused with this variable name. Have to guess what it does according to value assigned and check the usages. I think it can be named better, maybe refactor a bit.

packages/react-sdk/src/components/elements/block-arrow/Display.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+  const bodyTop = y + inset;
+  const bodyBottom = y + height - inset;
+  const bodyHeight = bodyBottom - bodyTop;
+  const centerY = y + height / 2;
+
+  const frameOffsetY = y - inset;
```
maheichyk: A bit confusing variable name, I think the term 'overlay' should be used here as well. Otherwise, it looks like it relates to frame positioning, which is not the case.

packages/react-sdk/src/components/elements/block-arrow/Display.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const frameHeight = height + inset * 2;
```
maheichyk: This variable is used later for the frame overlay. The value assigned is not correct. It shows the overlay bigger then the actual shape is. Please check.

packages/react-sdk/src/components/elements/block-arrow/Display.tsx
```diff
@@ -0,0 +1,153 @@
...
+    [x, bodyBottom],
+
+    // Close
+    [x, bodyTop],
+  ]
+    .map((pt) => pt.join(','))
+    .join(' ');
+
+  const textHeight = text ? Math.min(text.height, bodyHeight) : 0;
+  const textY = bodyTop + (bodyHeight - textHeight) / 2;
```
maheichyk: Looks strange that we recalculate text properties after they are already calculated by `getRenderProperties`. Please adjust the `getRenderProperties` to return all info that can be used for rendering.

`Math.min(text.height, bodyHeight)` to we really need this?

packages/react-sdk/src/components/elements/block-arrow/Display.tsx
```diff
@@ -0,0 +1,153 @@
...
+          active={active}
+          text={shape.text}
+          textAlignment={text.alignment}
+          textBold={text.bold}
+          textItalic={text.italic}
+          elementId={elementId}
+          x={text.position.x}
+          y={textY}
+          width={bodyWidth}
+          height={textHeight}
```
maheichyk: The issue here is that the right edge of the text is not really aligned to the arrow part. It currently ends somehow after the arrow part is started.

It also looks suboptimal to skip most of the space inside of the arrow and not to write the text here. I have checked the ticket and can see that it is expected to extend into the arrow as well. Please adjust the implementation.

maheichyk (changes requested):
I have left my comments, please have a look

packages/react-sdk/src/components/elements/block-arrow/Display.tsx
nurjinjafar: Good catch — using Math.max with a fixed pixel value can cause the arrowhead to exceed the shape width at small sizes. I’ll switch to a proportional clamp so the arrowhead width is always derived from (and bounded by) the total width.

packages/react-sdk/src/components/elements/block-arrow/Display.tsx
nurjinjafar: i agree with you the variable name is not optimal, i will rename it to something _selectionBoundingY_ or something like that since "_selectionBounding_.." is used other places in the code to refer to the selection rectangle.

packages/react-sdk/src/components/elements/block-arrow/Display.tsx
maheichyk: The frame overlay still doesn't look correct when it is over the frame, see screenshot:

block_arrow_overlay

packages/react-sdk/src/components/elements/block-arrow/Display.tsx
maheichyk: It is changed, but the width of the editable text is not accurate and needs proper calculation: 

block_arrow_text_position

packages/react-sdk/src/components/elements/block-arrow/getRenderProperties.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    // Assert text exists before checking properties
+    expect(view.text).toBeDefined();
+    expect(view.text!.position).toEqual({ x: 10, y: 27.5 });
+    expect(view.text!.width).toBe(80);
+    expect(view.text!.height).toBe(25);
+    expect(view.text!.alignment).toBe('center');
+    expect(view.text!.bold).toBe(false);
+    expect(view.text!.italic).toBe(false);
+    expect(view.text!.fontFamily).toBe('Inter');
```
maheichyk: Please use `toEqual` to compare all values and don't check them per value. Please update other similar changes.

Please check how it is done for other shapes.

packages/react-sdk/src/components/elements/block-arrow/getRenderProperties.ts
```diff
@@ -0,0 +1,58 @@
...
+    textAlignment,
+    textBold,
+    textItalic,
+    textSize,
+    textFontFamily,
+  } = shape;
+
+  // Padding inside the arrow body
+  const verticalPadding = height * 0.25; // space from top/bottom of arrow body
+  const horizontalPadding = 0;
```
maheichyk: Do we really need this variable with zero assigned?

packages/react-sdk/src/components/elements/block-arrow/getRenderProperties.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  const textHeight = height - verticalPadding * 2;
+  const textWidth = width * 0.8;
```
maheichyk: Need to calculate this properly so that the text has the correct width. Shouldn't be a magic number. 

See: https://github.com/nordeck/matrix-neoboard/pull/924#discussion_r2720172361

packages/react-sdk/src/components/BoardBar/pdf/createWhiteboardPdfElementShape.ts
```diff
@@ -106,3 +110,42 @@ function createElementShapeTriangle(element: ShapeElement): Content {
...
+
+  const points = [
+    { x: position.x, y: bodyTop }, // left-top of body
+    { x: position.x + bodyWidth, y: bodyTop }, // right-top of body
+    { x: position.x + bodyWidth, y: position.y }, // arrow head top
+    { x: position.x + width, y: centerY }, // tip
+    { x: position.x + bodyWidth, y: position.y + height }, // arrow head bottom
+    { x: position.x + bodyWidth, y: bodyBottom }, // right-bottom of body
+    { x: position.x, y: bodyBottom }, // left-bottom of body
+  ];
```
maheichyk: There is redundant logic between the PDF and Display that currently lacks test coverage. 

We should refactor this by moving the shared calculations into `getRenderProperties.ts`. 

This would allow to write a single suite of tests and reuse the logic in both places. This is a common pattern already. Please check polyline.

maheichyk (changes requested):

.changeset/two-parents-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@nordeck/matrix-neoboard-react-sdk': minor
+---
+
+add Block Arrow shape
```
maheichyk:
```suggestion
Add block arrow shape
```

packages/react-sdk/src/components/elements/block-arrow/getRenderProperties.ts
maheichyk: It looks similar to what requested in the ticket, but not really the same.

packages/react-sdk/src/components/BoardBar/pdf/createWhiteboardPdfElementShape.ts
```diff
@@ -106,3 +110,28 @@ function createElementShapeTriangle(element: ShapeElement): Content {
...
+
+function createElementShapeBlockArrow(element: ShapeElement): Content {
+  const { strokeColor, strokeWidth, text, points } =
+    getRenderBlockArrowProperties(element);
+
+  const isClosed =
+    points.length > 1 &&
+    points[0]?.x === points[points.length - 1]?.x &&
+    points[0]?.y === points[points.length - 1]?.y;
+  const pdfPoints = isClosed ? points.slice(0, -1) : points;
```
maheichyk: Why do we need this?

packages/react-sdk/src/components/BoardBar/pdf/createWhiteboardPdfElementShape.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      strokeWidth: 3,
+      text: '',
+    });
+
+    const { points } = getRenderBlockArrowProperties(element);
+    const isClosed =
+      points.length > 1 &&
+      points[0]?.x === points[points.length - 1]?.x &&
+      points[0]?.y === points[points.length - 1]?.y;
+    const pdfPoints = isClosed ? points.slice(0, -1) : points;
```
maheichyk: Why do we have this calculation repeated in the test?

packages/react-sdk/src/components/BoardBar/pdf/createWhiteboardPdfElementShape.test.ts
```diff
@@ -0,0 +1,77 @@
...
+describe('ellipse / circle', () => {
+  it.todo('renders ellipse for circle kind');
+  it.todo('renders ellipse for ellipse kind');
+});
+
+describe('triangle', () => {
+  it.todo('renders closed polyline triangle');
+});
+
+describe('createWhiteboardPdfElementShape', () => {
```
maheichyk: This test doesn't really needed. The PDF part is just a thin wrapper around `getRenderRectangleProperties`.

nurjinjafar: do u mean the whole test file?

packages/react-sdk/src/components/BoardBar/pdf/createWhiteboardPdfElementShape.test.ts
nurjinjafar: it is because with the current implementation of getRenderproperties i add a final point to close the polygon! But I searched, and apparently this step is redundant because SVG polygons close automatically! will change it

packages/react-sdk/src/components/BoardBar/pdf/createWhiteboardPdfElementShape.ts
nurjinjafar: check here https://github.com/nordeck/matrix-neoboard/pull/924/changes#diff-812287fb40ab526ef05593997f984d3c332a83321d66c47b175c67e2a711adc7R114-R131 i removed all calculations from createpdf file

packages/react-sdk/src/components/elements/block-arrow/Display.tsx
nurjinjafar: Bildschirmfoto 2026-02-05 um 09 45 38

packages/react-sdk/src/components/elements/block-arrow/Display.tsx
nurjinjafar: here now the new text box after the new calculations  Bildschirmfoto 2026-02-04 um 13 28 38

packages/react-sdk/src/components/elements/block-arrow/getRenderProperties.test.ts
nurjinjafar: check here ! https://github.com/nordeck/matrix-neoboard/pull/924/changes#diff-dddccec07d46de760f128c594a2c2fd04d4fd1f3e7346d9a21c162844f4e2a2eR60-R71

packages/react-sdk/src/components/elements/block-arrow/getRenderProperties.ts
nurjinjafar: the whole file is changed, please take another look!

packages/react-sdk/src/components/elements/block-arrow/getRenderProperties.ts
nurjinjafar: there is no magic number for the textboy anymore! check here: https://github.com/nordeck/matrix-neoboard/pull/924/changes#diff-0df104783d563a4d8260516c95e83642055ab354fbb3fb88f1dbe75e71f32f4dR85-R86

packages/react-sdk/src/components/BoardBar/pdf/createWhiteboardPdfElementShape.ts
nurjinjafar: is no longer used!

packages/react-sdk/src/components/BoardBar/pdf/createWhiteboardPdfElementShape.test.ts
nurjinjafar: test file is removed!

packages/react-sdk/src/components/elements/block-arrow/getRenderProperties.ts
```diff
@@ -0,0 +1,121 @@
...
+  const verticalTextPadding = bodyHeight > 40 ? 10 : 2;
+  const textTop = bodyTop + verticalTextPadding;
+  const textBottom = bodyBottom - verticalTextPadding;
+  const textHeight = Math.max(textBottom - textTop, 0);
+  const halfHeight = height / 2;
+
+  // Clamp utility to keep values inside bounds.
+  const clamp = (value: number, min: number, max: number) =>
+    Math.min(Math.max(value, min), max);
+
```
maheichyk, line 161: we already `clamp` in other places, please use the same function

packages/react-sdk/src/components/elements/block-arrow/getRenderProperties.ts
```diff
@@ -0,0 +1,121 @@
...
+  const headBoundaryAtY = (yPos: number) => {
+    if (halfHeight === 0) {
+      return x + bodyWidth;
+    }
+    const clampedY = clamp(yPos, y, y + height);
+    const distanceToTop = clampedY - y;
+    const distanceToBottom = y + height - clampedY;
+    const t = Math.min(distanceToTop, distanceToBottom) / halfHeight;
+    return bodyRight + arrowHeadWidth * t;
+  };
```
maheichyk: please move this into code into separate `function`

packages/react-sdk/src/components/elements/block-arrow/getRenderProperties.ts
nurjinjafar: done! https://github.com/nordeck/matrix-neoboard/pull/924/changes#diff-0df104783d563a4d8260516c95e83642055ab354fbb3fb88f1dbe75e71f32f4dR16-R17

packages/react-sdk/src/components/elements/block-arrow/getRenderProperties.ts
nurjinjafar: https://github.com/nordeck/matrix-neoboard/pull/924/changes#diff-0df104783d563a4d8260516c95e83642055ab354fbb3fb88f1dbe75e71f32f4dR132-R153

.changeset/two-parents-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@nordeck/matrix-neoboard-react-sdk': minor
+---
+
+Add block arrow shape
```
maheichyk, line 5: This PR only allows to create shape arrows to the right. It is very limited of course while the changeset sounds that to be more generic.

At lest I would expect that user can create a left and right block arrows that is also mentioned: https://github.com/nordeck/matrix-neoboard/pull/924#pullrequestreview-3678075742

packages/react-sdk/src/components/elements/block-arrow/getRenderProperties.ts
```diff
@@ -0,0 +1,153 @@
...
+}: HeadBoundaryArgs): number => {
+  if (halfHeight === 0) {
+    return x + bodyWidth;
+  }
+  const bodyRight = x + bodyWidth;
+  const clampedY = clamp(yPos, y, y + height);
+  const distanceToTop = clampedY - y;
+  const distanceToBottom = y + height - clampedY;
+  const t = Math.min(distanceToTop, distanceToBottom) / halfHeight;
+  return bodyRight + arrowHeadWidth * t;
```
maheichyk: Looks complicated to follow what is really does. I think this should be simplified. Maybe needs some better explicit tests.

This gets especially important as current PR doesn't currently support left arrow, bidirectional arrows.

packages/react-sdk/src/components/elements/block-arrow/getRenderProperties.ts
maheichyk: please change to function

.changeset/two-parents-shave.md
maheichyk: Agreed to work on it in a separate tickets.

packages/react-sdk/src/components/elements/block-arrow/getRenderProperties.ts
nurjinjafar: Bildschirmfoto 2026-02-09 um 10 02 41
 it is not in the scope of the ticket to support left arrow

packages/react-sdk/src/components/elements/block-arrow/getRenderProperties.ts
```diff
@@ -0,0 +1,100 @@
...
+  const bodyHeight = bodyBottom - bodyTop;
+  const bodyWidth = width - headWidth;
+
+  const bodyRightX = x + bodyWidth;
+
+  const verticalTextPadding = bodyHeight > 40 ? 10 : 2;
+
+  const textHeight = Math.max(bodyHeight - verticalTextPadding * 2, 0);
+  const textWidth = Math.max(
+    bodyWidth + headWidth * (1 - bodyHeight / height) - horizontalPadding * 2,
```
maheichyk: This `textWidth` formula is not correct and only "works" currently because `bodyHeight / height` ratio is `0.5`.

I think that this code should be either generic (work with different ratios) which is much preferred. Or explicit that it is not.

segnord: If you are talking about this part `headWidth * (1 - bodyHeight / height)`, I've verified that it works for any ratio, not just 0.5

maheichyk: Did a code refactoring to simplify calculations to be able to follow how it actually renders without spending to much time reasoning the code. Basically based on your changes to split original calculations.

maheichyk (dismissed):

packages/react-sdk/src/components/elements/block-arrow/Display.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const boundingRect = calculateBoundingRectForPoints(points);
+
+  const renderedChild = (
+    <g data-testid={dataTestid}>
+      {/* Block arrow shape with narrow body */}
+      <polygon
+        points={points.map(({ x, y }) => `${x},${y}`).join(' ')}
+        fill={shape.fillColor}
+        stroke={strokeColor}
+        strokeWidth={strokeWidth}
```
mgcm, line 55: Is this shape missing the `data-connect-type={"connectable-element"}` property? this seems to be required for maintaining shape connections when resizing shapes  (in `findConnectData`) but I don't see any issues by not having it... @maheichyk  ?

packages/react-sdk/src/components/Whiteboard/Draft/calculateShapeCoords.test.ts
```diff
@@ -71,6 +72,15 @@ describe('calculateShapeCoords', () => {
         },
         endCoords: { x: 750, y: 725 },
       },
+      'block-arrow',
```
mgcm, line 77: this is wrong and within the ellipse shape data.

```suggestion
    ],
    [
      'block-arrow',
```

mgcm (changes requested):
left a couple of comments - also `block-arrow` should be added to the [crdt documentation](https://github.com/nordeck/matrix-neoboard/blob/15973408a41609ff90b5e4c7092305a7c129e332/docs/model/crdt-documents.md?plain=1#L132) as a possible shape kind.

packages/react-sdk/src/components/elements/block-arrow/Display.tsx
maheichyk: Yes, added. It is used when connecting to stop activation of elements that are below of the block arrow.

packages/react-sdk/src/components/Whiteboard/Draft/calculateShapeCoords.test.ts
maheichyk: Thanks applied

mgcm (approved):

PR #951, opened 2026-03-16
Implement single element rotation

segnord:
Implements rotation for single elements: shapes, images. Allows to select, move, resize rotated elements together with other elements. Rotated elements can be connected with lines.

rotation1

rotation2


.changeset/seven-meals-sink.md
```diff
@@ -0,0 +1,5 @@
+---
+'@nordeck/matrix-neoboard-react-sdk': minor
+---
+
+Added object rotation feature (experiental)
```
maheichyk:
```suggestion
Add element rotation
```
If we keep a feature flag we should maybe mention it as well.

packages/react-sdk/src/components/ElementOverridesProvider/utils.ts
```diff
@@ -15,12 +15,23 @@
...
 ): T {
+  if (isRotateableElement(element)) {
+    return {
+      ...element,
+      height: override?.height ?? element.height,
+      width: override?.width ?? element.width,
+      position: override?.position ?? element.position,
+      rotation: override?.rotation ?? element.rotation,
+    };
+  }
```
maheichyk: It now handles the shapes and images but there is the code below in this method that does handles them without rotation. So with this change that another code will not be used anymore. We should improve that. We also should add tests for this method.

packages/react-sdk/src/components/elements/image/ImageDisplay.tsx
```diff
@@ -183,11 +185,14 @@ function ImageDisplay({
...
   ) : null;
 
+  const rot = rotation ?? 0;
+
   const renderedChild =
     imageUri !== undefined && !loadError ? (
       <Image
         data-testid={`element-${elementId}-image`}
         href={imageUri}
+        transform={`rotate(${rot} ${position.x + width / 2} ${position.y + height / 2})`}
```
maheichyk: This always adds a `transform` event when rotation is `undefined`. Don't see why we should add this attribute if no rotation or rotation is changed to `0`

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Connectable/ConnectableElement.tsx
```diff
@@ -75,6 +77,23 @@ export function ConnectableElement({
     [x, y, height, width, kind],
   );
 
+  const rotatedPoints = points.map((p) => {
+    if (rotation && p) {
+      const point = {
+        x: p[0],
+        y: p[1],
+      };
```
maheichyk: I think we can improve that, we don't need to map over the points at all if no rotation or 0. No need to check if `p` is defined - it is according to types. It also looks like we maybe should refactor a bit to use a `Point` when the points are defined to have less or no helper code to convert from array to `Point`.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/ResizeElement.tsx
```diff
@@ -308,11 +312,30 @@ export function ResizeElement({ elementIds }: ResizeElementProps) {
...
+      const rot = element.rotation ?? 0;
+      const center = {
+        x: element.width / 2,
+        y: element.height / 2,
+      };
+      return `translate(${offsetX} ${offsetY}) rotate(${rot} ${center.x} ${center.y})`;
+    }
+
+    return `translate(${offsetX} ${offsetY})`;
+  };
```
maheichyk: Don't see why `transform` needs to be an arrow function.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/ResizeElement.tsx
```diff
@@ -308,11 +312,30 @@ export function ResizeElement({ elementIds }: ResizeElementProps) {
...
+
+  const rotation = () => {
+    if (elementIds.length === 1 && isRotateableElement(activeElements[0])) {
+      const element = activeElements[0];
+      return element.rotation;
+    }
+    return 0;
+  };
```
maheichyk: Similar don't see why `rotation` is an arrow function. Also using `0` instead of `undefined` doesn't look reasonable.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/ResizeHandle.tsx
```diff
@@ -182,6 +185,10 @@ export function ResizeHandle(props: ResizeHandleProps) {
     scale,
   });
 
+  const getRotatedCursor = () => {
+    return rotateCursor(cursor, rotation ?? 0);
+  };
+
```
maheichyk: No need to use an arrow function.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
```diff
@@ -39,6 +50,55 @@ export function isLineElementHandlePosition(
   return handlePosition.name === 'start' || handlePosition.name === 'end';
 }
 
+export function computeDragWithRotation(
+  event: DragEvent,
+  elements: Elements,
+  viewportWidth: number,
+  viewportHeight: number,
+  gridCellSize?: number,
+) {
```
maheichyk: Add a return type for the function. Especially for ones that are exported. Otherwise it gets much more difficult to read the code. Please check other places similar.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
```diff
@@ -39,6 +50,55 @@ export function isLineElementHandlePosition(
...
+          ? rawDragX
+          : snapToGrid(rawDragX, gridCellSize);
+      const dragY =
+        gridCellSize === undefined
+          ? rawDragY
+          : snapToGrid(rawDragY, gridCellSize);
+
+      // user tries to rotate a single element, the resize borders are also rotated
+      // but this function will return drag deltas as if the object wasn't rotated
+      return { dragX, dragY };
```
maheichyk: It doesn't look reasonable to rely on unrotated coordinates to check if a shape is within whiteboard bound during a single shape drag. It easily goes outside of the whiteboard when rotated element is dragged.

It should work similarly to how it already works in this PR when several shapes (including rotated ones) are moved to the edge of the whiteboard. They do not go outside of it.

In addition this code added also copies code from `computeDrag` (and calls it as well...) I wonder if we should have these two functions or what we should refactor to look better.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
```diff
@@ -117,25 +177,43 @@ export function calculateDragDimension(
...
 
+export function getLockAspectRatio(
+  event: DragEvent,
+  elements: Elements,
+  invertLockAspectRatio: boolean = false,
+) {
+  const hasRotated = hasRotatedElements(elements);
+  const multiselect = checkMultiselect(elements);
+
+  if (hasRotated && multiselect) return true;
```
maheichyk: I think we should add a comment why we force aspect ratio in this case as it looks strange.

However looking at other changes it probably should be possible to update the PR that lock is not required in this case. However haven't tried that.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
```diff
@@ -117,25 +177,43 @@ export function calculateDragDimension(
...
+
+  return invertLockAspectRatio ? !event.lockAspectRatio : event.lockAspectRatio;
+}
+
 export function calculateDimensions(
   handlePositionName: HandlePositionName,
+  elementsPossiblyRotated: Elements,
```
maheichyk: Need to check.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
```diff
@@ -308,8 +577,9 @@ export function computeResizing(
   }
 
   if (isLineElementHandlePosition(handlePosition)) {
-    const { dragX, dragY } = computeDrag(
+    const { dragX, dragY } = computeDragWithRotation(
```
maheichyk: Why do we use a compute drag logic for possibly **rotatable elements** when a line (non rotatable) handle is used?

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
```diff
@@ -342,67 +612,81 @@ export function computeResizing(
...
+    pointResizer: (p: Point) => Point;
+  }[] = Object.entries(resizableProperties.elements).map(
+    ([elementId, element]) => {
+      const { override: unrotatedOverride, pointResizerUnrotated } =
+        resizeInfoUnrotated(
+          elementId,
+          element,
+          resizableProperties,
+          dimensions,
+        );
```
maheichyk: This doesn't look to be in a correct place: it is executed without any need if a single rotated element is rotated. See the code below.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
```diff
@@ -342,67 +612,81 @@ export function computeResizing(
...
+
+      // Note: in multiselect mode resizing is always performed along the SVG axis
+      // and not along the rotated element's axis, thus we skip it.
+
+      if (isRotateableElement(element) && !multiselect && element.rotation) {
```
maheichyk: I found this confusing that we handle this logic in a loop when it gets applied when a collection has a single element. Maybe could be improved.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
```diff
@@ -479,25 +766,42 @@ function computeResizingConnectingPathElement(
...
+        resizers &&
+        resizers[connectedElementId] &&
+        isRotateableElement(elementWithConnectorPoints) &&
+        elementWithConnectorPoints.rotation
+      ) {
+        const resizeFn = resizers[connectedElementId];
+        const resizedPoint = resizeFn({ x: pointX, y: pointY });
+        drag.x = resizedPoint.x;
+        drag.y = resizedPoint.y;
+      }
```
maheichyk: `resizers` (not sure if it is a good name) are computed for both: unrotated and rotated cases. However, they are used only when the element is rotated. It doesn't look like they are needed for the unrotated case. If they are still needed, I'm not sure we really need to distinguish between these cases since `rotation` param could be used to handle the logic differently.

In the case of rotated the purpose is clear but it takes much time to understand the code while reading. I think we would simplify/improve this part.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateElement.tsx
```diff
@@ -0,0 +1,256 @@
...
+  elementId: string;
+};
+
+export type Rotation = {
+  angle: number;
+};
+
+export type RotateProperties = Rotation & {
+  element: ElementBase;
+};
```
maheichyk: Not used

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateElement.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+export function RotateElement({ elementId }: RotateElementProps) {
```
maheichyk, line 40: Needs an `index.ts` for `Rotatable` package, see: https://github.com/nordeck/matrix-widget-toolkit/blob/main/docs/adrs/adr005-widget-and-module-export-structure.md

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateElement.tsx
```diff
@@ -0,0 +1,256 @@
...
+  const setElementOverride = useSetElementOverride();
+  const elements = useElementOverrides([elementId]);
+  const activeElements = Object.values(elements);
+  const { viewportWidth, viewportHeight } = useSvgCanvasContext();
```
maheichyk: Is deprecated

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateElement.tsx
```diff
@@ -0,0 +1,256 @@
...
+    const updates = [update];
+
+    // update the connected arrows
+    updates.push(...rotateConnectedPaths(event));
+
+    setElementOverride(updates);
+    setElementOverrideUpdates(updates);
+  };
+
+  function rotateConnectedPaths(event: DragEvent) {
```
maheichyk: Can we move this function `rotateConnectedPaths` outside of component to `utils.ts` similar to `Resizable`, `Movable`

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateHandler.tsx
```diff
@@ -0,0 +1,246 @@
...
+  handlePosition: RotatorHandlePosition;
+  onDrag?: Dispatch<DragEvent>;
+  onDragStart?: Dispatch<DragEvent>;
+  onDragStop?: Dispatch<DragEvent>;
+};
+
+export type DragEvent = {
+  newAngle: number;
+  referenceAngle: number;
+  lockAspectRatio: boolean;
```
maheichyk: it is not about aspect ratio lock

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateHandler.tsx
```diff
@@ -0,0 +1,246 @@
...
+};
+
+export function RotateHandler(props: RotateHandleProps) {
+  const { scale } = useSvgScaleContext();
+  const nodeRef = useRef<SVGRectElement>(null);
+  const { calculateSvgCoords } = useSvgCanvasContext();
+
+  const { handlePosition: position, onDrag, onDragStart, onDragStop } = props;
```
maheichyk: Can avoid props variable and destruct inside of function parameters.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateHandler.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+    const handle = {
+      x: rotationInitialPosition.x,
+      y: rotationInitialPosition.y,
+    };
+
+    const cursor = calculateSvgCoords({
+      x: clamp(data.x * scale, 0, viewportWidth),
+      y: clamp(data.y * scale, 0, viewportHeight),
+    });
```
maheichyk: data.x, data.y are coordinates on screen, not position in whiteboard, clamp logic is not correct here

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateHandler.tsx
```diff
@@ -0,0 +1,246 @@
...
+    if (dragEvent) {
+      dragEvent({
+        newAngle: angleInteger,
+        referenceAngle: rotationInitialPosition.rot,
+        lockAspectRatio: event.shiftKey,
+      });
+    }
+  };
+
+  const handleStart = (event: DraggableEvent, data: DraggableData) => {
```
maheichyk: I don't see `useCallback` used constantly in this component. While it may be not needed, we do use it in all other components. Should be consistent across file. Should be consistent also across code in my opinion.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateHandler.tsx
```diff
@@ -0,0 +1,246 @@
...
+  const handleStop = (event: DraggableEvent, data: DraggableData) => {
+    setRotationInitialPosition(undefined);
+    dispatchDragEvent(onDragStop, event, data);
+  };
+
+  const theme = useTheme();
+  const selectionAnchorCornerRadius = 3 / scale;
+  const selectionBorderWidth = 2 / scale;
+
+  const getRotatedHandleCorner = () => {
```
maheichyk: A plain function no need to rerender each time.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateHandler.tsx
```diff
@@ -0,0 +1,246 @@
...
+        ></rect>
+      </DraggableCore>
+    </g>
+  );
+}
+
+export type HandlePos = {
+  containerWidth: number;
+  containerHeight: number;
+};
```
maheichyk: Not used

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/rotatorMath.ts
```diff
@@ -0,0 +1,188 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { clamp } from 'lodash';
+import { Element, Elements, Point } from '../../../../state';
+import { isRotateableElement } from '../../../../state/crdt/documents/elements';
+
+export const hasRotatedElements = (elements: Elements) =>
```
maheichyk: Used once in Resizable, not in Rotatable, should think if should be really there and not in resize in how the PR is currently look like. Otherwise would need to export it properly and it looks like too much work for this function.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/rotatorMath.ts
```diff
@@ -0,0 +1,188 @@
...
+  Object.entries(elements)
+    .map(([_, element]) => !!(isRotateableElement(element) && element.rotation))
+    .includes(true);
+
+export const checkMultiselect = (elements: Elements) =>
```
maheichyk: Similar used once in Resizable, also very simple logic I would think that it does more to be called a multiselect, not sure if really needed.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/rotatorMath.ts
```diff
@@ -0,0 +1,188 @@
...
+  Object.entries(elements).length > 1;
+
+export function rotateCursor(cursor: string, angle: number): string {
```
maheichyk: used once in ResizeHandle.tsx

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/rotatorMath.ts
```diff
@@ -0,0 +1,188 @@
...
+  ];
+  const nangle = clampAngle(angle);
+  const times = Math.round(nangle / 45);
+  const idx = cursors.findIndex((s) => cursor === s);
+  if (idx === -1) return cursor;
+  const res = cursors[(idx + times) % 8];
+  return res;
+}
+
+export function calculateBoundaryWithRotationHandleForElements(
```
maheichyk: Does not too much and used once in ElementBarWrapper.tsx

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/rotatorMath.ts
```diff
@@ -0,0 +1,188 @@
...
+  if (elements.length !== 1) return null;
+  if (!isRotateableElement(elements[0])) return null;
+  return calculateBoundaryWithRotationHandle(elements[0], scale);
+}
+
+export function getMinMaxFromPoints(
+  points: Point[],
+  rotation: number,
+  center: Point,
+) {
```
maheichyk: No function return type.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/rotatorMath.ts
```diff
@@ -0,0 +1,188 @@
...
+  const minY = rotated
+    .map((o) => o.y)
+    .reduce((prev, cur) => (cur <= prev ? cur : prev));
+
+  return {
+    min: { x: minX, y: minY },
+    max: { x: maxX, y: maxY },
+    width: Math.abs(maxX - minX),
+    height: Math.abs(maxY - minY),
+  };
```
maheichyk: Isn't that a `calculateBoundingRectForPoints` over rotated points?

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/rotatorMath.ts
```diff
@@ -0,0 +1,188 @@
...
+export function calculateBoundaryWithRotationHandle(
+  element: Element,
+  scale: number,
+) {
+  if (!isRotateableElement(element)) {
+    throw Error(`Element can't be rotated and can't have a rotation handle`);
+  }
+
+  // element has rotation
+  const x = element?.position.x ?? 0;
```
maheichyk: element is defined here

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Selection/ElementBarWrapper.tsx
```diff
@@ -126,6 +146,35 @@ const InfiniteElementBarWrapper: React.FC<ElementBarWrapperProps> = ({
   const offsetOnDiv = 10;
   const { scale, transformPointSvgToContainer } = useSvgScaleContext();
 
+  const getBoundariesWhenRotatedInContainer = (
```
maheichyk: Does that need to be an arrow function?

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Selection/ElementBorder.tsx
```diff
@@ -83,93 +84,110 @@ export function ElementBorder({ elementIds, padding = 1 }: ElementBorderProps) {
     elements[0].kind === 'line' &&
     getRenderProperties(elements[0]);
 
+  const transform = () => {
```
maheichyk: arrow function?

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Selection/ElementOutline.tsx
```diff
@@ -26,35 +27,46 @@ export function ElementOutline({ elementIds }: ElementOutlineProps) {
...
         } else {
           width = element.width;
           height = element.height;
         }
 
+        let transform = '';
+
+        // rotated object
+        if (isRotateableElement(element)) {
+          const rot = element.rotation ?? 0;
```
maheichyk: similar could avoid `0` assignment if `undefined`

packages/react-sdk/src/state/crdt/documents/elements.ts
```diff
@@ -348,6 +397,12 @@ export function isShapeElementPair(
   return pair[1].type === 'shape';
 }
 
+export function isRotateableElement(
```
maheichyk: Add test

maheichyk (changes requested):

packages/react-sdk/src/state/crdt/documents/elements.ts
```diff
@@ -77,6 +78,7 @@ export type ShapeElement = ElementBase & {
   textFontFamily: TextFontFamily;
   connectedPaths?: string[];
   attachedFrame?: string;
+  rotation?: number;
```
maheichyk, line 81: Need to update schema validation

.changeset/seven-meals-sink.md
segnord: I  will remove the feature flag

packages/react-sdk/src/components/ElementOverridesProvider/utils.ts
segnord: Almost everywhere where there are such checks if the rotation is 0 or undefined, it will execute the code that existed before the rotation was introduced to minimize the possibility of breaking existing slides

packages/react-sdk/src/components/elements/image/ImageDisplay.tsx
segnord: ok, I will fix it

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/ResizeElement.tsx
segnord: I will see if only the "rotation" section can be optionally rendered

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Connectable/ConnectableElement.tsx
segnord: Migrated the array to a Point object

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Selection/ElementBarWrapper.tsx
segnord: Simplified in the new commit, but was useful in a function format during dev

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Selection/ElementBorder.tsx
segnord: moved to a util mudule

packages/react-sdk/src/state/crdt/documents/elements.ts
segnord: Added tests

packages/react-sdk/src/state/crdt/documents/elements.ts
segnord: Schema and tests added

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/ResizeElement.tsx
segnord: Most of the stuff that were arrow functions was so it was more flexible to add console logs and return values conditionally without using ?:

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
segnord: Added return types

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
segnord: I think it's a bit hard to change because it deals with the user's "feeling" when resizing a rotated object, maybe this clamping can be improved in a follow up PR, the original compute drag is there to process the unrotated logic without going through rotation code

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
segnord: added comment

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
segnord: it's mostly  because  if there's even 1 rotated object it has to go through rotated boundary calculation, using this array we can detect the rotated object, but if you pass an empty, then every object is calculated as unrotated (the original logic)

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
segnord: Fixed

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
segnord: Fixed

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
segnord: I remember there was weird resize behaviours if the conditions were different during the dev, so I added conditions until it looked right for the user. Maybe in the end it can be simplified

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
segnord: I simplified it to use the resizerFn regardless of object rotation, because this function already knows how to resize any point in it's own element. The rotation check is redundant here, I will keep the original code just in case we don't get a resizerFn for each object in this function and will assume the original unrotated calculations.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateElement.tsx
segnord: Fixed

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateElement.tsx
segnord: Added index,ts

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateElement.tsx
segnord: Fixed

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateElement.tsx
segnord: Fnction extracted, also added test

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateHandler.tsx
segnord: Renamed to angleSnap

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateHandler.tsx
segnord: Fixed

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateHandler.tsx
segnord: Removed clamp here

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateHandler.tsx
segnord: Added useCallback to drag handlers

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateHandler.tsx
segnord: I don't think this function will cause a rerender by being executed

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/RotateHandler.tsx
segnord: Fixed

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/rotatorMath.ts
segnord: I tried to put stuff that has something to do  with rotation into the rotation module, it just happened that it may have been useful only in resize modules

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/rotatorMath.ts
segnord: Fixed

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/rotatorMath.ts
segnord: Yes, it looks very similar to `calculateBoundingRectForPoints` but I didn't want to calculate a boundary, but extreme points, also `Math.abs` to handle negative coordinates width/height (more for debug purposes, mostly interested in the points). Maybe calculateBoundingRectForPoints can be used instead, but I don't know it is worth changing

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/rotatorMath.ts
segnord: Fixed

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Selection/ElementOutline.tsx
segnord: Fixed

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/ResizeHandle.tsx
segnord: Fixed

packages/react-sdk/src/components/ElementOverridesProvider/utils.ts
segnord: There's also the "frame" element that goes to the 'non-path' branch

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/rotatorMath.ts
segnord: I will remove this function

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/rotatorMath.ts
maheichyk: Has refactored this.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.rotation.test.ts
```diff
@@ -0,0 +1,503 @@
...
+  computeDragWithRotation,
+  computeResizing,
+  computeResizingConnectingPathElement,
+  getLockAspectRatio,
+  pointResizerRotatedFactory,
+  pointResizerUnrotatedFactory,
+  resizeInfoRotated,
+} from './utils';
+
+describe('resizing utils for rotated elements', () => {
```
maheichyk: Moved this utils.rotation.test.ts content into utils.ts

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
```diff
@@ -39,6 +50,55 @@ export function isLineElementHandlePosition(
...
+        cursor,
+        {
+          x: element.position.x + element.width / 2,
+          y: element.position.y + element.height / 2,
+        },
+        -angle,
+      );
+
+      const rawDragX = clamp(unrotatedDragEventPoint.x, 0, viewportWidth);
+      const rawDragY = clamp(unrotatedDragEventPoint.y, 0, viewportHeight);
```
maheichyk: Clamping here doesn't work as might be expected for a rotated element. Basically it is removed. We need to consider how to make element not to go outside of svg canvas when rotated element is resized.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
```diff
@@ -117,25 +177,45 @@ export function calculateDragDimension(
...
+  const multiselect = checkMultiselect(elements);
+
+  // The aspect ratio is locked because the rotated element in multi-select
+  // is being resized at an angle and it will look weird with unlocked aspect ratio.
+  if (hasRotated && multiselect) return true;
+
+  return invertLockAspectRatio ? !event.lockAspectRatio : event.lockAspectRatio;
+}
+
 export function calculateDimensions(
```
maheichyk, line 191: Did quite a lot of refactoring into this util.ts file to make sure the rotation logic fits better to existing code. Also refactoring changes are related to existing code in this file, not only to rotation logic.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
```diff
@@ -294,6 +376,227 @@ function computeResizingOfLine(
...
+/**
+ * Factory for a point rotating functions when resizing a rotated element
+ *  This function will predict where a point on this object will be
+ * after applying the resize (in SVG coordinates)
+ * it unrotates the point, applies the "unrotated resize logic" and
+ * rotates the point over the new center, applying the error shift.
+ * @param options
+ * @returns function that applies the object's current resize transformation to any point
+ */
+export function pointResizerRotatedFactory(options: {
```
maheichyk: Did refactorings related to these factories

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
```diff
@@ -294,6 +376,227 @@ function computeResizingOfLine(
...
+    override,
+    pointResizerUnrotated,
+  };
+}
+
+/**
+ * For a rotated element.
+ * Returns override after resizing and a resizer function specific to the perfomed resize action
+ */
+export function resizeInfoRotated(
```
maheichyk: Changed to handle both roated and not rotated elements.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Resizable/utils.ts
```diff
@@ -342,67 +645,82 @@ export function computeResizing(
...
+        elementId,
+      };
+    },
+  );
+
+  // just collect all the resizer functions into a map elementId->resizerFn
+  const resizers: Record<string, (p: Point) => Point> = {};
+  for (const o of resizeInfo) {
+    if (o.pointResizer) resizers[o.elementId] = o.pointResizer;
+  }
```
maheichyk: This resizer is not needed as a record. Basically we need a resizer (point resizer) when not rotated and single resizer function for rotated case. So not need to have a function for each element. Refactored.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Rotatable/rotatorMath.ts
```diff
@@ -0,0 +1,220 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { clamp } from 'lodash';
+import { Element, Elements, Point } from '../../../../state';
+import { isRotateableElement } from '../../../../state/crdt/documents/elements';
+
+export function hasRotatedElements(elements: Elements): boolean {
```
maheichyk: This file is imported and used in many places, its code is splitted to different files where the logic is need and also moved to some existing generic utils.

maheichyk:
Refactored addressing the issues in the code while reading, reviewing, testing the PR.

segnord:
Please also merge the latest `main` branch or rebase

maheichyk (approved):

PR #963, opened 2026-04-20
Migrate slides into frames

maheichyk:
When application runs in infinite canvas mode and the user has opened a whiteboard with slides, he is asked to update to frames:

update1

If a stored frames document is opened in slides user is asked to upgrade:

update_2a

Export and import are updated to use the version of whiteboard document.

This PR introduces a separate whiteboard document version for frames.


packages/react-sdk/src/locales/en/neoboard.json
```diff
@@ -318,6 +318,12 @@
     "title": "Slide Overview",
     "title_frame": "Frame Overview"
   },
+  "slidesMigrationDialog": {
+    "content": "Your existing slides will be migrated to frames. To enable migration, please download a backup of your whiteboard content first.",
+    "migrate": "Migrate",
+    "migrationFailed": "Failed to migrate the slides",
+    "title": "Migrate slides to frames"
+  },
```
maheichyk, line 326: Maybe could/should adjust and needs German translations.

maheichyk: @HarHarLinks  could you please have a look

HarHarLinks: What do you think about this, it is a bit outside the box with the idea to try not to confuse users with technobabble:
```suggestion
  "slidesMigrationDialog": {
    "content": "Your existing slides will be automatically upgraded to the new frames. Please download a backup of your whiteboard content before you continue.",
    "migrate": "Continue",
    "migrationFailed": "Failed to upgrade slides to frames",
    "title": "Upgrade Slides to Frames"
  },
```
(assuming we use title case consistently - https://www.titlecase.com/)

maheichyk: Thanks, I like the suggestion.

packages/react-sdk/src/locales/de/neoboard.json
```diff
@@ -318,6 +318,12 @@
     "title": "Folienübersicht",
     "title_frame": "Rahmenübersicht"
   },
+  "slidesMigrationDialog": {
+    "content": "",
+    "migrate": "",
+    "migrationFailed": "",
+    "title": ""
```
HarHarLinks:
```suggestion
    "content": "Deine vorhandenen Folien werden automatisch aktualisiert, um die neuen Rahmen zu nutzen. Bitte lade ein Backup deiner Whiteboard-Inhalte herunter bevor du fortfährst.",
    "migrate": "Fortfahren",
    "migrationFailed": "Aktualisierung von Folien zu Rahmen fehlgeschlagen",
    "title": "Folien zu Rahmen aktualisieren"
```

packages/react-sdk/src/locales/en/neoboard.json
mgcm: `migration` and `upgrade` still sound a bit like technobabble.

How about:

title: "Slides are now Frames"

content: "Your existing Slides will now be shown as Frames -- a new way to organize your content.

Please download a copy of your whiteboard before continuing."

migrationFailed:  "An issue occurred when moving Slides into Frames".

What do you think?

packages/react-sdk/src/components/SlidesMigration/SlidesMigration.tsx
```diff
@@ -0,0 +1,116 @@
...
+      });
+    } catch (err) {
+      logger.error('Cannot migrate slides to to frames', err);
+      setIsError(true);
+    } finally {
+      setIsLoading(false);
+      setIsMigrated(true);
+    }
+
+    whiteboardInstance.clearUndoManager();
```
mgcm: do we want to clear the undo manager even if the migration fails? I think this should be in the `try` block instead.

packages/react-sdk/src/components/SlidesMigration/SlidesMigration.tsx
```diff
@@ -0,0 +1,116 @@
...
+    setIsLoading(true);
+    try {
+      whiteboardInstance.transformSlidesToFrames({
+        whiteboardWidth,
+        whiteboardHeight,
+        frameWidth,
+        frameHeight,
+      });
+    } catch (err) {
+      logger.error('Cannot migrate slides to to frames', err);
```
mgcm:
```suggestion
      logger.error('Cannot migrate slides to frames', err);
```

packages/react-sdk/src/components/SlidesMigration/SlidesMigrationDialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+};
+
+export function SlidesMigrationDialog({
+  open,
+  onMigrate,
+  isMigrationDisabled,
+  isLoading,
+  isError,
+  additionalButtons,
+}: SlideMigrationDialogProps) {
```
mgcm:
```suggestion
}: SlidesMigrationDialogProps) {
```
for coherence with component naming.

packages/react-sdk/src/components/SlidesMigration/SlidesMigration.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  const handleMigration = useCallback(async () => {
+    setIsLoading(true);
+    try {
+      whiteboardInstance.transformSlidesToFrames({
+        whiteboardWidth,
+        whiteboardHeight,
+        frameWidth,
+        frameHeight,
+      });
```
mgcm: if we add `setOpen(false)` here, we would not need the `isMigrated` state variable, right?

mgcm (changes requested):

packages/react-sdk/src/locales/en/neoboard.json
HarHarLinks: yes this is why I tried to remove `migrate` from the wording, assuming `upgrade` is more familiar. but if it can be avoided altogether that also sounds good to me.

maheichyk: Changed the texts, based on what Milton has suggested. Anyway the '--' looks a bit strange to me :). Haven't changed German translations yet. Let's have a look how it is now and change if needed.

Now a whiteboard document has a different version for frames and the dialog is updated to say that the user has to upgrade (NeoBoard) to work with this document.

packages/react-sdk/src/components/SlidesMigration/SlidesMigration.tsx
maheichyk: The `open` idea was to use it to represent the state: count of slides and now a presense of the mismatched snapshot. Didn't want to mix with the migration itself been triggered.

packages/react-sdk/src/components/SlidesMigration/SlidesMigration.tsx
maheichyk: remove this statement, don't think we need it in the changed implementation

packages/react-sdk/src/locales/en/neoboard.json
```diff
@@ -318,6 +318,14 @@
     "title": "Slide Overview",
     "title_frame": "Frame Overview"
   },
+  "slidesMigrationDialog": {
+    "content": "Your existing Slides will now be shown as Frames - a new way to organize your content. Please download a copy of your whiteboard before continuing.",
+    "contentUpgradeRequired": "A newer version is required to work with the stored document.",
+    "migrationFailed": "An issue occurred when moving Slides into Frames",
+    "title": "Slides are now Frames",
+    "titleUpgradeRequired": "Upgrade is Required",
```
mgcm:
```suggestion
    "titleUpgradeRequired": "Upgrade required",
```

packages/react-sdk/src/locales/en/neoboard.json
```diff
@@ -318,6 +318,14 @@
     "title": "Slide Overview",
     "title_frame": "Frame Overview"
   },
+  "slidesMigrationDialog": {
+    "content": "Your existing Slides will now be shown as Frames - a new way to organize your content. Please download a copy of your whiteboard before continuing.",
+    "contentUpgradeRequired": "A newer version is required to work with the stored document.",
```
mgcm:
```suggestion
    "contentUpgradeRequired": "This document requires a more recent version of Neoboard that uses Frames instead of Slides.",
```

packages/react-sdk/src/state/export/whiteboardDocumentExport.ts
```diff
@@ -99,33 +102,92 @@ const whiteboardFileExportSchema = Joi.object<WhiteboardFileExport, true>({
...
+export function whiteboardDocumentVersionToExportString(
+  whiteboardDocumentVersion: WhiteboardDocumentVersion,
+): string {
+  return `net.nordeck.whiteboard@v${parseInt(whiteboardDocumentVersion) + 1}`;
+}
+
+export function extractWhiteboardDocumentVersionFromExportString(
+  value: string,
+): WhiteboardDocumentVersion {
+  const match = value.match(/net.nordeck.whiteboard@v([\d.]+)/);
```
mgcm:
```suggestion
  const match = value.match(/^net\.nordeck\.whiteboard@v([\d]+)$/);
```

I get that this is only extracting the version number but the current regex is way too flexible -it will match stuff like `xyz netXnordeckYneoboard@1.2.3 foobar` and still return version 1 - which is not wrong but not quite right? 🗡️ 

My proposed change matches the exact `net.nordeck.whiteboard@v` string prefix and nothing else and also restricts valid version numbers to integers (and also restricts suffixes).

packages/react-sdk/src/state/crdt/documents/whiteboardDocument.ts
```diff
@@ -39,12 +48,13 @@ export type WhiteboardDocument = {
   slideIds: YArray<string>;
 };
 
+// Generate an initial slide. All whiteboards will have this slide id!
+const slideId = 'IN4h74suMiIAK4AVMAdl_';
```
mgcm, line 52: this constant is defined here but also defined as `slide0` and used in 7 test files - should we not make this globally accessible and with the same name?

packages/react-sdk/src/state/synchronizedDocumentImpl.test.ts
```diff
@@ -911,24 +1080,45 @@ describe('SynchronizedDocumentImpl', () => {
...
-function createExampleDocument(): Document<Example> {
-  return YDocument.create<Example>(exampleMigrations, '0');
+const exampleMigrations = createMigrations([initMigration], '0');
+const example1Migrations = createMigrations(
+  [initMigration, migrateToList],
+  '1',
+);
+
+function createExampleDocument(
+  exampleDocumentVersion: '0' | '1' = '0',
```
mgcm, line 1104: Overall, I think there is some semantic confusion of using 0 and 1 as internal document versions, which then get written as v1 and v2 in the actual data, and then are enumerated as `.Initial` and `.Frames` in other places...

I don't want to add further effort into this PR but is there any strong reason not to use just '1' and '2' as the document versions across the codebase? The fact that v2 adds proper frame support is something that belongs in the documentation and code logic + comments.

Also, how to handle these code paths when we have 5 document versions?

If we have:

```
export enum WhiteboardDocumentVersion {
v1 = '1',
v2 = '2',
}
```

and

```
export type WhiteboardDocumentVersionType = '1' | '2';
```

and use these instead across the codebase, which then becomes:

```exampleDocumentVersion: WhiteboardDocumentVersionType = WhiteboardDocumentVersion.v1, ```

a bit more verbose but less confusing, no?

**Review by mgcm — COMMENTED** [review]
Added some additional comments on the code, which overall is fine.

Regarding functionality, I have two points:

1. With this PR, every single board, even if it was already created with Infinite Canvas and Frames enabled, will be prompted for migration. This is mostly an issue for us, in our stage and demo environments but still a bit annoying and maybe we can do that silently?

2. These changes allow for a user that supports v2 to open a board that was created by a user that only supports v1, and still perform the migration, thus making the board unreadable to the creator. This is a roles / permissions issue, and again, maybe not a use case we should support, but still worth considering if we accept this.

File: packages/react-sdk/src/locales/en/neoboard.json (hunk 8 above)
Comment (HarHarLinks) [reply packages/react-sdk/src/locales/en/neoboard.json]: Branding: *NeoBoard

Also, I think "whiteboard" is clearer than "document".

```suggestion
    "contentUpgradeRequired": "This whiteboard requires a more recent version of NeoBoard that uses Frames instead of Slides.",
```

packages/react-sdk/src/locales/en/neoboard.json
```diff
@@ -318,6 +318,14 @@
...
+    "migrationFailed": "An issue occurred when moving Slides into Frames",
```
HarHarLinks: this seems like we're leaving the user hanging in a broken state without a hint of what to do next. is it possible to do better?

packages/react-sdk/src/locales/de/neoboard.json
```diff
@@ -318,6 +318,14 @@
     "title": "Folienübersicht",
     "title_frame": "Rahmenübersicht"
   },
+  "slidesMigrationDialog": {
+    "content": "",
+    "contentUpgradeRequired": "",
+    "migrationFailed": "",
+    "title": "",
+    "titleUpgradeRequired": "",
+    "update": ""
```
HarHarLinks:
```suggestion
    "content": "Deine vorhandenen Folien werden nun als Rahmen dargestellt - eine neue Art, deine Inhalte zu strukturieren. Bitte lade ein Backup deines Whiteboards herunter bevor du fortfährst.",
    "contentUpgradeRequired": "Dieses Whiteboard erfordert eine aktuellere Version von NeoBoard, das Rahmen statt Folien unterstützt.",
    "migrationFailed": "Aktualisierung von Folien zu Rahmen fehlgeschlagen",
    "title": "Folien sind jetzt Rahmen",
    "titleUpgradeRequired": "Aktualisierung erforderlich",
    "update": "Fortfahren"
```

packages/react-sdk/src/locales/en/neoboard.json
maheichyk: Generic error handling here for completeness. Not sure if we can/should do it better.

packages/react-sdk/src/state/export/loadWhiteboardFromExport.ts
```diff
@@ -53,12 +51,6 @@ export function generateLoadWhiteboardFromExport(
       const [addSlide, slideId] = generateAddSlide();
       addSlide(doc);
 
-      if (isInfiniteCanvasMode()) {
-        const setSlideFrameElementIds =
-          generateSetSlideFrameElementIds(slideId);
-        setSlideFrameElementIds(doc);
-      }
```
maheichyk, line 60: Was incorrect to remove, breaks import for documents with frames version, properly changed in 93a0282fe78d915d5b84a3366d241be096c9cb0d

packages/react-sdk/src/locales/en/neoboard.json
maheichyk: Latest suggestion from HarHarLinks is applied

packages/react-sdk/src/state/crdt/documents/whiteboardDocument.ts
maheichyk: I think the original idea of that was to not export `slideId` from `whiteboardDocument.ts` so that it is clear that client code is not using it (for instance to perform custom changes in documents). 

However it is useful in the tests obviously. Thought about that for some time while working on this PR, didn't want to export it, maybe should move to some common code used for testing, maybe `lib/testUtils/whiteboardDocument.ts` and use only in tests? No strong opinion.

packages/react-sdk/src/state/synchronizedDocumentImpl.test.ts
maheichyk: We already have `'0'` used as a document version of Yjs document stored, however we also have `net.nordeck.whiteboard@v1` as version of the document exported/imported.

I thought that it would be useful to have the version named, for example: `initial`, `frames`. Because keeping versions named in the form: v0, v1 (or v1, v2) would be harder to use, we also have other objects that are named for the specific version in this PR like migrations or schemas, using a generic `vX` format here maybe would be harder to follow.

But finding right word for version name may be not that straightforward, the `vX` format maybe would make tests to sound more easy to understand. For example `'v0'` (or `'v1'`) would be better because a `frame` word could be confused with frame element and we have many tests like that.

It looks like we can use
```
export enum WhiteboardDocumentVersion {
v0 = '0',
v1 = '1',
}
```

and 

```
export type WhiteboardDocumentVersionType = '0' | '1';
```

To be zero based to avoid any confusing. When exported it should be +1 with `v0` exported to `'net.nordeck.whiteboard@v1'`.

maheichyk:
Thanks for the review.

>     1. With this PR, every single board, even if it was already created with Infinite Canvas and Frames enabled, will be prompted for migration. This is mostly an issue for us, in our stage and demo environments but still a bit annoying and maybe we can do that silently?

See the point, addressed in: 0281da842e854dc2e148e7bc4f36743349972b5b

>     2. These changes allow for a user that supports v2 to open a board that was created by a user that only supports v1, and still perform the migration, thus making the board unreadable to the creator. This is a roles / permissions issue, and again, maybe not a use case we should support, but still worth considering if we accept this.

Agreed that currently it is not really an issue, we may think about that later, maybe allow only moderators to do the update.

packages/react-sdk/src/locales/en/neoboard.json
maheichyk: Hm, however the suggested change is explicitly about Frames. But actually the original idea was more generic: the user have to use a new version of NeoBoard, maybe Frames, maybe something else. We are basically receiving an unrecognized whiteboard document version and don't know what it contains.

HarHarLinks: I suppose that's fair. Should we also add generic messages as a fallback and only show these messages about slides and frames when the version mismatch is exactly v1/v2? Even though we don't really have an upgrade situation in this case, but more a feature flag situation.

packages/react-sdk/src/state/crdt/documents/whiteboardDocument.ts
mgcm: How about we rename this one to `firstSlideId` and keep it local. For the tests, do as you say: move `slide0` declaration to `lib/testUtils/whiteboardDocument.ts` and rename it to `firstSlideId` and then import it in the tests that need it ?

packages/react-sdk/src/locales/en/neoboard.json
maheichyk: I would not try to differentiate these cases really, just more efforts with maybe not so much value. A generic message to update without mentioning slides or frames would be good in my opinion.

packages/react-sdk/src/state/synchronizedDocumentImpl.test.ts
maheichyk: Updated 868649422fba2aa5d0a86538e74109508e88d459

packages/react-sdk/src/state/crdt/documents/whiteboardDocument.ts
maheichyk: Thought about it again. Overall I feel fine to change that. However, the changes in this PR simply follow the existing approach. I would prefer not to refactor existing existing approaches unless necessary, as the diff will grow even further and make changes hard to follow. Suggest to make it in separate PR.

packages/react-sdk/src/locales/en/neoboard.json
mgcm: I'm happy with the current state.

packages/react-sdk/src/state/synchronizedDocumentImpl.test.ts
mgcm: thanks, this is better 👍

packages/react-sdk/src/state/crdt/documents/whiteboardDocument.ts
mgcm: agreed!

mgcm (approved):

PR #977, opened 2026-05-20
fix undo redo shortcuts

nurjinjafar:
**Problem**
                                                                                                                                                                                             
  The undo/redo shortcuts _(Ctrl+Z, Meta+Z, Ctrl+Shift+Z, Ctrl+Y)_ were broken for users on non-QWERTY keyboard layouts (e.g. German QWERTZ).                                                  
   
  `react-hotkeys-hook` matches hotkeys by `e.code` _(physical key position)_ by default. 
On a QWERTZ keyboard, the `Z` and `Y `keys are physically swapped compared to QWERTY, so:                     
                                                                                                                                                                                           
  - Pressing `Ctrl+Z` (the Z key) fires the `ctrl+y` (redo) handler — because the German Z key sits at the QWERTY Y position (KeyY)                                                              
  - Pressing `Ctrl+Y` (the Y key) fires the `ctrl+z` (undo) handler — because the German Y key sits at the QWERTY Z position (KeyZ)                                                            
                                                                                                                                                                                             
  An attempt to fix this by adding `useKey: true` to the `meta+z`/`ctrl+z` registration. 
This option switches matching to `e.key` (the produced character), which is correct for                 
  layout-independence. However, `useKey: true` has a https://github.com/JohannesKlauss/react-hotkeys-hook/issues/1318: the modifier keys in the hotkey string are completely ignored, so `ctrl+z`
   with `useKey: true` fired on any bare `z` keypress without requiring `Ctrl` or `Meta`.                                                                                                            
                                                                                                                                                                                           
 **Solution**

  Register the bare key ('z' / 'y') with `useKey: true` — which is honest about what the library does — and check the required modifier (e.metaKey on macOS, e.ctrlKey elsewhere) manually     
  inside the callback. 

This is a workaround until the upstream fix (https://github.com/JohannesKlauss/react-hotkeys-hook/pull/1298 /
  https://github.com/JohannesKlauss/react-hotkeys-hook/pull/1317) is merged into the library.                                                                                                
                                                                                                                                                                                           
  The same fix is applied to the `ctrl+y` (redo) handler to prevent it double-firing on QWERTZ when the user presses their Z key (see                                                          
  https://github.com/JohannesKlauss/react-hotkeys-hook/issues/1239).
                                                                                                                                                        
https://github.com/user-attachments/assets/367e2019-728c-4b21-9af7-1584bbbcca9e



<!--- Please include the following in your Pull Request when applicable: -->


.changeset/upset-towns-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@nordeck/matrix-neoboard-react-sdk': patch
+---
+
+fix undo redo shortcuts for non querty keyboard users
```
maheichyk: `querty` ?

packages/react-sdk/src/components/Shortcuts/UndoRedoShortcuts/UndoRedoShortcuts.tsx
```diff
@@ -15,39 +15,53 @@
...
       }
-
-      whiteboardManager.getActiveWhiteboardInstance()?.undo();
+      // this is a workaround for non qwerty keyboards
+      // see issue here: https://github.com/PostHog/code/issues/1797
+      if (e.shiftKey) {
+        whiteboardInstance.redo();
+      } else {
+        whiteboardInstance.undo();
+      }
```
maheichyk: Is it a workaround? The link is misleading here: it doesn't mention the changes that we do in this PR. Maybe there is a better link to the issue of `react-hotkeys-hook` that explains the problem. It may be better just to write few words in comments about changes in this PR.

packages/react-sdk/src/components/Shortcuts/UndoRedoShortcuts/UndoRedoShortcuts.tsx
```diff
@@ -15,39 +15,53 @@
...
+    },
+    {
+      scopes: HOTKEY_SCOPE_WHITEBOARD,
+      enableOnContentEditable: true,
+      useKey: true,
```
maheichyk, line 52: A `useKey` option is added in this PR for this listener. The modifiers are not handled as might be expected, here are the issues:

- Pressing 'z' applies undo without **ANY** modifier.
- Pressing 'meta+Z_QWERTY_KEY' on non non-QWERTY keyboard will **call the listener**. This is especially unexpected for German keyboard.

Please check `react-hotkeys-hook` documentation and existing issues and verify the changes that are proposed.

packages/react-sdk/src/components/Shortcuts/UndoRedoShortcuts/UndoRedoShortcuts.tsx
```diff
@@ -15,39 +15,53 @@
...
     },
-    { scopes: HOTKEY_SCOPE_WHITEBOARD, enableOnContentEditable: true },
-    [isViewingPresentation, whiteboardManager],
   );
 
   useHotkeys(
-    isMacOS() ? ['meta+shift+z'] : ['ctrl+shift+z', 'ctrl+y'],
+    'ctrl+y',
```
maheichyk: This separate redo listener has the same issue this PR is addressing: using a non-QWERTY layout fires this listener only when QWERTY key is pressed

maheichyk (changes requested):
Left the comments to the issues found. Also, the PR description is misleading: it links to the issue in another project saying that the library we are using doesn't support non-QWERTY layouts and the same time states that a `useKey: true` options is used to address the issue. Let's rephrase that to describe what the state is and what we change in the PR.

.changeset/upset-towns-yawn.md
nurjinjafar: fixed the typo! thx for noticing!

packages/react-sdk/src/components/Shortcuts/UndoRedoShortcuts/UndoRedoShortcuts.tsx
nurjinjafar: Bildschirmfoto 2026-06-18 um 12 09 12

packages/react-sdk/src/components/Shortcuts/UndoRedoShortcuts/UndoRedoShortcuts.tsx
nurjinjafar: thanks for noticing! the handler now, checks if the z key is pressed, if it has a modifier or not and later if it has a shift or not to detect the undo or redo

packages/react-sdk/src/components/Shortcuts/UndoRedoShortcuts/UndoRedoShortcuts.tsx
nurjinjafar: also is fixed! the same done with the undo is done in here!
Bildschirmfoto 2026-06-18 um 12 28 45

maheichyk (approved):
Looks good, thanks

PR #988, opened 2026-06-08
fix: explicitly track and unsubscribe per-peer-connection subscription

Marcel-Nordeck:
handleSessionJoined creates two RxJS subscriptions (observeMessages, observeStatistics) per peer connection without storing them. While WebRtcPeerConnection.close() completes both subjects (causing auto-unsubscription in normal flow), if destroy() is called before a peer cleanly disconnects the subscriptions are orphaned.

### DIFF
--- a/.changeset/slick-memes-hammer.md
+++ b/.changeset/slick-memes-hammer.md
@@ -1,0 +1,5 @@
+---
+'@nordeck/matrix-neoboard-react-sdk': patch
+---
+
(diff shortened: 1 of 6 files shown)


Marcel-Nordeck:
This probably needs some pair testing. I am not sure how to best test this. In theory this works. In practice I have no good way of confirming it doesnt cause side effects @mgcm maybe could also take a look as the one who last(?) worked on this code?

mgcm:
I went thru your suggested changes and consider the subscription mapping not necessary - indeed what was missing was applying the `takeUntil` and actually closing the peer connections, which do complete those subjects on destroy.

I also checked the mRTC path and fixed leaking the peer connections across reconnects. 

Also added tests for both issues and a changeset.

PR #989, opened 2026-06-08
perf: debounce statistics emission to reduce GC pressure

Marcel-Nordeck:
Two sources were generating excessive short-lived allocations:

1. webRtcCommunicationChannel: addPeerConnectionStatistics called cloneDeep(statistics) once per peer connection per second. With N peers that's N deep clones/sec. Route all stat mutations through a pendingStatistics Subject debounced at 200ms — now 1 clone/sec max regardless of peer count.

2. yDocument.observeStatistics: JSON.stringify(getRoot().toJSON()) ran on every Yjs document change. During active collaboration this fires many times per second on potentially large documents. Add debounceTime (1000ms) before the map — stats are display-only so the delay is fine.

Update yDocument.test.ts to use fake timers for the statistics test.

---

This makes neoboard standalone use a third of the memory footprint for me

### DIFF
--- a/packages/react-sdk/src/state/communication/webRtcCommunicationChannel.ts
+++ b/packages/react-sdk/src/state/communication/webRtcCommunicationChannel.ts
@@ -22,6 +22,7 @@ import {
   NEVER,
   Observable,
   Subject,
+  debounceTime,
   distinctUntilChanged,
   filter,
   firstValueFrom,
@@ -59,6 +60,10 @@ export class WebRtcCommunicationChannel implements CommunicationChannel {
   private readonly statistics: CommunicationChannelStatistics =
     emptyCommunicationChannelStatistics();
   private readonly peerConnections: PeerConnection[] = [];
+  // Internal trigger for statistics emission. Callers push to this subject;
+  // emissions are debounced so that rapid bursts (e.g. one tick per peer
+  // connection per second) produce only one cloneDeep + emit per window.
+  private readonly pendingStatistics = new Subject<void>();
   private turnServers: TurnServer | undefined;
 
   constructor(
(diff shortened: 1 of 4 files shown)


segnord (commented):
I tested this locally and it does seem to work as expected, the statistics dialog updates every second instead of immediately. But I'd like @mgcm to also look at it

packages/react-sdk/src/state/communication/webRtcCommunicationChannel.ts
```diff
@@ -59,6 +60,10 @@ export class WebRtcCommunicationChannel implements CommunicationChannel {
   private readonly statistics: CommunicationChannelStatistics =
     emptyCommunicationChannelStatistics();
   private readonly peerConnections: PeerConnection[] = [];
+  // Internal trigger for statistics emission. Callers push to this subject;
+  // emissions are debounced so that rapid bursts (e.g. one tick per peer
+  // connection per second) produce only one cloneDeep + emit per window.
+  private readonly pendingStatistics = new Subject<void>();
```
maheichyk, line 66:
```suggestion
  private readonly pendingStatisticsSubject = new Subject<void>();
```
To keep it consistent with this and other files

packages/react-sdk/src/state/crdt/y/yDocument.ts
```diff
@@ -147,18 +147,21 @@ export class YDocument<
...
-        };
-      }),
+    return concat(
+      // Emit once immediately so subscribers receive a populated baseline
+      // before any change fires (preserves the ordering that synchronizedDocumentImpl
+      // relies on: contentSizeInBytes is set before snapshotOutstanding changes).
+      of(null).pipe(map(compute)),
+      // Debounce subsequent changes so that rapid bursts (e.g. during
+      // collaboration) produce only one JSON.stringify per idle window.
+      this.changesSubject.pipe(debounceTime(1000), map(compute)),
```
maheichyk, line 164: I have found the `debounceTime` to be confusingly used here, since it delays events and waits for no event emission during the `dueTime`. Looks like `trottleTime` should be used instead. Yes there is no hard dependency on this behavior at the moment, but this PR changes the current behavior in this regard. Maybe could use `throttleTime`, what do you think?

packages/react-sdk/src/state/communication/webRtcCommunicationChannel.ts
```diff
@@ -73,6 +78,14 @@ export class WebRtcCommunicationChannel implements CommunicationChannel {
   ) {
     this.logger.log('Creating communication channel');
 
+    // Batch rapid statistics updates (e.g. one tick per peer per second) into
+    // a single cloneDeep + emit per 200 ms window.
+    this.pendingStatistics
+      .pipe(debounceTime(200), takeUntil(this.destroySubject))
```
maheichyk, line 84: Looks like `throttleTime` should be better alternative to `debounceTime` - we will not need to wait for no event during `dueTime`.

PR #990, opened 2026-06-08
perf: reduce selection overhead with Set lookups and pointer throttle

Marcel-Nordeck:
Replace O(n) includes() checks with O(1) Set.has() in WhiteboardHost, change getActiveElementIds() from O(n×m) to O(n+m) via Set filtering, and cap drag-select pointer events at ~60fps via a Date.now() time-gate to avoid redundant intersection calculations.

### DIFF
--- a/packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Selection/DragSelect.tsx
+++ b/packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Selection/DragSelect.tsx
@@ -15,7 +15,14 @@
  */
 
 import { styled, useTheme } from '@mui/material';
-import { PointerEvent, useCallback, useEffect, useMemo, useState } from 'react';
+import {
+  PointerEvent,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
 import { filterRecord } from '../../../../lib';
 import {
   Point,
@@ -124,10 +131,13 @@ export function DragSelect() {
     setDragSelectStartCoords();
   }, [setDragSelectStartCoords]);
 
+  const lastMoveRef = useRef(0);
   const handlePointerMove = useCallback(
     (event: PointerEvent<SVGRectElement>) => {
-      const point = calculateSvgCoords({ x: event.clientX, y: event.clientY });
-      setEndCoords(point);
+      const now = Date.now();
+      if (now - lastMoveRef.current < 16) return;
+      lastMoveRef.current = now;
+      setEndCoords(calculateSvgCoords({ x: event.clientX, y: event.clientY }));
     },
     [calculateSvgCoords],
   );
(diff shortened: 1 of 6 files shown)


packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Selection/DragSelect.tsx
```diff
@@ -124,10 +131,13 @@ export function DragSelect() {
     setDragSelectStartCoords();
   }, [setDragSelectStartCoords]);
 
+  const lastMoveRef = useRef(0);
   const handlePointerMove = useCallback(
     (event: PointerEvent<SVGRectElement>) => {
-      const point = calculateSvgCoords({ x: event.clientX, y: event.clientY });
-      setEndCoords(point);
+      const now = Date.now();
+      if (now - lastMoveRef.current < 16) return;
```
segnord: I wonder if the throttling should be time based, or coordinate based, so that if it's the same coordinate or small jitter we don't proceed. Also it would be better to make 16 into a constant so it has a description that it's 60 updates per second maybe in a form of `1000/fps`.

packages/react-sdk/src/components/Whiteboard/WhiteboardHost.tsx
```diff
@@ -74,6 +74,10 @@ const WhiteboardHost = ({
   const { isShowCollaboratorsCursors, dragSelectStartCoords } =
     useLayoutState();
   const { activeElementIds } = useActiveElements();
+  const activeElementSet = useMemo(
```
segnord: I wonder if the set of Ids can improve performance elsewhere and we can return it as a 2nd object from `useActiveElements`

segnord (dismissed):
Looks ok, I only had minor comments

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Selection/DragSelect.tsx
Marcel-Nordeck: Done the constant in [bab5430](https://github.com/nordeck/matrix-neoboard/pull/990/commits/bab5430accc4b0ba9998c3067f7c213182c1e3ff)

packages/react-sdk/src/components/Whiteboard/WhiteboardHost.tsx
Marcel-Nordeck: Done that in [bab5430](https://github.com/nordeck/matrix-neoboard/pull/990/commits/bab5430accc4b0ba9998c3067f7c213182c1e3ff). Not sure where else to use it but it is now available.

packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Selection/DragSelect.tsx
Marcel-Nordeck: I think the coordinate makes no sense. This tries to debounce based on frame rate (to feel smooth) while coordinate would just smooth just mouse jitter essentially. So the 2 things solve different problems I think

segnord (approved):

PR #994, opened 2026-06-10
fix(presentation): atomically capture edit mode before updating state

Marcel-Nordeck:
When a PRESENT_FRAME message was received, isEditMode was read to decide whether to clear the undo manager, then updated separately. Out-of-order messages from the network could cause the transition check to fire at the wrong time, silently clearing a user's undo history.

Fix: capture the previous value before calling setEditMode() so the transition check compares old vs new state atomically.

### DIFF
--- a/packages/react-sdk/src/state/presentationManagerImpl.ts
+++ b/packages/react-sdk/src/state/presentationManagerImpl.ts
@@ -140,12 +140,13 @@ export class PresentationManagerImpl implements PresentationManager {
             if (content.view) {
               this.setCurrentPresenter(senderSessionId);
 
-              // clear the undo manager unless the user is already in edit mode
-              if (!this.isEditMode && content.view.isEditMode) {
+              const wasEditMode = this.isEditMode;
+              this.setEditMode(content.view.isEditMode);
+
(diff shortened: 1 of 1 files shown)


packages/react-sdk/src/state/presentationManagerImpl.ts
```diff
@@ -140,12 +140,13 @@ export class PresentationManagerImpl implements PresentationManager {
...
-              if (!this.isEditMode && content.view.isEditMode) {
+              const wasEditMode = this.isEditMode;
+              this.setEditMode(content.view.isEditMode);
+
+              // clear the undo manager only when transitioning into edit mode
+              if (!wasEditMode && content.view.isEditMode) {
                 whiteboardInstance.clearUndoManager();
               }
-
-              this.setEditMode(content.view.isEditMode);
```
maheichyk, line 148: It doesn't look to be the difference from the original implementation: `this.setEditMode(content.view.isEditMode)` was just moved to another line and its position doesn't influence any other statements, the `wasEditMode` is simply assigned from `content.view.isEditMode)` (before it was the same but without variable). Could you check if something is missing?

PR #995, opened 2026-06-11
perf(react-sdk): stabilize context provider values with useMemo

Marcel-Nordeck:
Five context providers were building their value objects inline, causing all context consumers to re-render on every provider render regardless of whether the data actually changed. Wrap each provider value in useMemo so consumer subtrees only re-render when the relevant state changes.

Also stabilizes showImportWhiteboardDialog with useCallback so the ImportWhiteboardDialogContext value reference is truly stable.

Providers fixed:
- ElementAttachFrameProvider (setter context)
- ImportWhiteboardDialogProvider
- ImageUploadProvider
- SnackbarProvider
- ConnectionStateProvider

### DIFF
--- a/packages/react-sdk/src/components/ElementAttachFrameProvider/ElementAttachFrameProvider.tsx
+++ b/packages/react-sdk/src/components/ElementAttachFrameProvider/ElementAttachFrameProvider.tsx
@@ -96,14 +96,21 @@ export function ElementAttachFrameProvider({
     [elementAttachFrame, attachedElementsMovedByFrame, connectingPathIds],
   );
 
+  const setterContext = useMemo<ElementAttachFrameSetterContextType>(
+    () => ({
+      setElementAttachFrame,
+      setAttachedElementsMovedByFrame,
+      setConnectingPathIds,
+    }),
+    [
+      setElementAttachFrame,
+      setAttachedElementsMovedByFrame,
+      setConnectingPathIds,
+    ],
+  );
+
   return (
-    <ElementAttachFrameSetterContext.Provider
-      value={{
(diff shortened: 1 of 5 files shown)


packages/react-sdk/src/components/connection-state/ConnectionStateProvider.tsx
```diff
@@ -249,16 +249,25 @@ export const ConnectionStateProvider: React.FC<PropsWithChildren> = function ({
     setSnapshotLoadDialogOpen(false);
   }, [setSnapshotLoadDialogOpen]);
 
+  const ctx = useMemo(
```
maheichyk, line 252: Could you add changeset?

packages/react-sdk/src/components/ElementAttachFrameProvider/ElementAttachFrameProvider.tsx
```diff
@@ -96,14 +96,21 @@ export function ElementAttachFrameProvider({
     [elementAttachFrame, attachedElementsMovedByFrame, connectingPathIds],
   );
 
+  const setterContext = useMemo<ElementAttachFrameSetterContextType>(
```
maheichyk, line 99: Unfortunately it doesn't fix the issue that all elements are re-rendered when element is moved. Should be fixed in other PR.

maheichyk: Actually only Moveable elements are rerendered, not display, but anyway an issue, fix: https://github.com/nordeck/matrix-neoboard/pull/1012

PR #996, opened 2026-06-11
perf(react-sdk): stabilize setTextToolsEnabled via ref in TextEditor

Marcel-Nordeck:
setTextToolsEnabled is a prop with a default of () => {} - a new function reference on every render. Including it in useEffect and useCallback deps caused those hooks to re-run on every render regardless of actual changes.

Switch to the ref-for-event-handler pattern: always write the latest prop into a ref and call through the ref in effects and callbacks, so none of them need setTextToolsEnabled in their deps array.

### DIFF
--- a/packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Text/TextEditor.tsx
+++ b/packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Text/TextEditor.tsx
@@ -132,14 +132,17 @@ export function TextEditor({
   const [isEditMode, setEditMode] = useState(editModeOnMount);
   usePauseHotkeysScope(HOTKEY_SCOPE_WHITEBOARD, isEditMode && editable);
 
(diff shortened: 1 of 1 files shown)


packages/react-sdk/src/components/Whiteboard/ElementBehaviors/Text/TextEditor.tsx
```diff
@@ -132,14 +132,17 @@ export function TextEditor({
   const [isEditMode, setEditMode] = useState(editModeOnMount);
   usePauseHotkeysScope(HOTKEY_SCOPE_WHITEBOARD, isEditMode && editable);
 
+  const setTextToolsEnabledRef = useRef(setTextToolsEnabled);
+  setTextToolsEnabledRef.current = setTextToolsEnabled;
```
maheichyk, line 136: Could you please add changeset?

PR #997, opened 2026-06-11
perf(react-sdk): lazy state initializer and stable shade picker key

Marcel-Nordeck:
useDistinctObserveBehaviorSubject: wrap subject.getValue() in an arrow function so React only calls it on mount, not on every render.

ShadePicker: use the shade hex value as the list key instead of the array index, so React correctly reconciles items if the shade list ever reorders.


packages/react-sdk/src/state/useDistinctObserveBehaviorSubject.ts
```diff
@@ -27,7 +27,7 @@ import { ObservableBehaviorSubject } from './types';
 export function useDistinctObserveBehaviorSubject<T>(
   subject: ObservableBehaviorSubject<T>,
 ): T {
-  const [value, setValue] = useState(subject.getValue());
+  const [value, setValue] = useState(() => subject.getValue());
```
maheichyk, line 30: Could you please add changeset?

PR #1022, opened 2026-08-06
Remove Widget API metadata from the OpenID Token sent to lk-jwt service

mgcm:
Since [version 0.4.0](https://github.com/element-hq/lk-jwt-service/releases/tag/v0.4.0), the LiveKit JWT Service does a [stricter validation](https://github.com/element-hq/lk-jwt-service/commit/6573de0d93a8a50dd9a07ba2c4c0ac806d7ad813#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R357) of the OpenID Token passed down to `/sfu/get`.

Because the Widget API carries around the Identity Verification request permission state and returns that on calls to `requestOpenIDConnectToken`, we need to strip the state out.

### DIFF
--- a/packages/react-sdk/src/state/communication/discovery/autodiscovery.ts
+++ b/packages/react-sdk/src/state/communication/discovery/autodiscovery.ts
@@ -1,7 +1,7 @@
 /*
  * Copyright 2018 New Vector Ltd
  * Copyright 2019 The Matrix.org Foundation C.I.C.
- * Copyright 2025 Nordeck IT + Consulting GmbH
+ * Copyright 2025-2026 Nordeck IT + Consulting GmbH
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


packages/react-sdk/src/state/communication/discovery/autodiscovery.ts
```diff
@@ -178,7 +178,15 @@ export default class AutoDiscovery {
...
-          openid_token: openIDToken,
+          // Only forward the credential fields. The Widget API result carries
+          // identity request permissions metadata such as `state` that the
+          // LiveKit JWT Token Service rejects.
+          openid_token: {
+            access_token: openIDToken.access_token,
+            expires_in: openIDToken.expires_in,
+            matrix_server_name: openIDToken.matrix_server_name,
+            token_type: openIDToken.token_type,
+          },
```
maheichyk, line 189: Fixes the issue. Removes `state`, `original_request_id` that produce the error.

However I wonder If we should apply this at this place where jwt service applies validation.

These `state`, `original_request_id` are not relevant for openIdToken and are some artifacts that matrix widget api returns unfortunately.

Should we instead remove them as soon as we get a response from widget api? Right after the call to `widgetApi.requestOpenIDConnectToken`?

A better idea could fix it in matrix widget toolkit that this is not passed to the widget at all.

Anyway I am fine with the changes. Feel free to update it or merge as it is.

maheichyk (approved):

PR #1027, opened 2026-08-19
"Line thickness" control added for selected polylines in the ElementBar

segnord:

The last changed value will be used when adding a new polyline in the current browser session, the default line thickness is 4. The path rendering tech didn't change, the feature was already implemented using the strokeWidth field of the path element with the hardcoded value of 4. Changes only apply to elements of type = "path" and kind = "polylines" (added by the Pen tool)

### DIFF
--- a/.changeset/true-snails-retire.md
+++ b/.changeset/true-snails-retire.md
@@ -1,0 +1,5 @@
+---
+'@nordeck/matrix-neoboard-react-sdk': minor
+---
+
+"Line thickness" control added for selected polyline elements in the Element Bar. The last changed thickness value will be used when adding a new polyline in the current browser session, the default line thickness is 4.
--- a/packages/react-sdk/src/components/ElementBar/PolylineThicknessSelect/PolylineThicknessSelect.test.tsx
+++ b/packages/react-sdk/src/components/ElementBar/PolylineThicknessSelect/PolylineThicknessSelect.test.tsx
@@ -1,0 +1,157 @@
+/*
+ * Copyright 2026 Nordeck IT + Consulting GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { MockedWidgetApi, mockWidgetApi } from '@matrix-widget-toolkit/testing';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import axe from 'axe-core';
+import { ComponentType, PropsWithChildren } from 'react';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import {
+  WhiteboardTestingContextProvider,
+  mockEllipseElement,
+  mockPolylineElement,
+  mockWhiteboardManager,
+} from '../../../lib/testUtils/documentTestUtils';
+import { WhiteboardSlideInstance } from '../../../state';
+import { Toolbar } from '../../common/Toolbar';
+import { LayoutStateProvider } from '../../Layout';
+import { defaultStrokeWidth } from '../../Whiteboard/constants';
+import { PolylineThicknessSelect } from './PolylineThicknessSelect';
+
+let widgetApi: MockedWidgetApi;
+
+afterEach(() => widgetApi.stop());
+
+beforeEach(() => {
+  widgetApi = mockWidgetApi();
+});
+
+describe('<PolylineThicknessSelect/>', () => {
+  let Wrapper: ComponentType<PropsWithChildren<{}>>;
+  let slide: WhiteboardSlideInstance;
+
+  beforeEach(() => {
+    const { whiteboardManager } = mockWhiteboardManager({
+      slides: [
+        [
+          'slide-0',
+          [
+            ['element-0', mockPolylineElement({ strokeWidth: 8 })],
+            ['element-7', mockPolylineElement({ strokeWidth: 7 })],
+            [
+              'element-undefined',
+              mockPolylineElement({ strokeWidth: undefined }),
+            ],
+            ['ellipse', mockEllipseElement()],
+          ],
+        ],
+      ],
+    });
+    slide = whiteboardManager
+      .getActiveWhiteboardInstance()!
+      .getSlide('slide-0');
+    slide.setActiveElementIds(['element-0']);
+
+    Wrapper = ({ children }) => (
+      <WhiteboardTestingContextProvider
+        whiteboardManager={whiteboardManager}
+        widgetApi={widgetApi}
+      >
+        <LayoutStateProvider>
+          <Toolbar>{children}</Toolbar>
+        </LayoutStateProvider>
+      </WhiteboardTestingContextProvider>
+    );
+  });
+
+  it('should render without exploding', async () => {
+    render(<PolylineThicknessSelect />, { wrapper: Wrapper });
+
+    const select = screen.getByRole('combobox', {
+      name: 'Select Line Thickness',
+    });
+
+    expect(select).toBeInTheDocument();
+  });
+
+  it('should have no accessibility violations', async () => {
+    const { container } = render(<PolylineThicknessSelect />, {
+      wrapper: Wrapper,
+    });
+
+    expect(
+      screen.getByRole('combobox', { name: 'Select Line Thickness' }),
+    ).toBeInTheDocument();
+
+    expect(await axe.run(container)).toHaveNoViolations();
+  });
+
+  it('should show the stroke width of the active element', async () => {
+    slide.setActiveElementId('element-7');
+    render(<PolylineThicknessSelect />, { wrapper: Wrapper });
+
+    const select = screen.getByRole('combobox', {
+      name: 'Select Line Thickness',
+    });
+
+    expect(select).toHaveTextContent('7');
+  });
+
+  it('should show the default stroke width if the active element has none', async () => {
+    slide.setActiveElementId('element-undefined');
+    render(<PolylineThicknessSelect />, { wrapper: Wrapper });
+
+    const select = screen.getByRole('combobox', {
+      name: 'Select Line Thickness',
+    });
+
+    expect(select).toHaveTextContent('4');
+  });
+
+  it('should show the stroke width of the first selected element if several elements are active', async () => {
+    slide.setActiveElementIds(['element-7', 'element-0']);
+    render(<PolylineThicknessSelect />, { wrapper: Wrapper });
+
+    const select = screen.getByRole('combobox', {
+      name: 'Select Line Thickness',
+    });
+
+    expect(select).toHaveTextContent('7');
+  });
+
+  it('should always contain the default stroke width as an option', async () => {
+    slide.setActiveElementId('element-7');
+    render(<PolylineThicknessSelect />, { wrapper: Wrapper });
+
+    await userEvent.click(
+      screen.getByRole('combobox', { name: 'Select Line Thickness' }),
+    );
+
+    expect(
+      screen.getByRole('option', { name: `${defaultStrokeWidth}` }),
+    ).toBeInTheDocument();
+  });
+
+  it('should not render if the active element is not a polyline', async () => {
+    slide.setActiveElementId('ellipse');
+    render(<PolylineThicknessSelect />, { wrapper: Wrapper });
+
+    expect(
+      screen.queryByRole('combobox', { name: 'Select Line Thickness' }),
+    ).not.toBeInTheDocument();
+  });
+});
--- a/packages/react-sdk/src/components/ElementBar/PolylineThicknessSelect/PolylineThicknessSelect.tsx
+++ b/packages/react-sdk/src/components/ElementBar/PolylineThicknessSelect/PolylineThicknessSelect.tsx
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2026 Nordeck IT + Consulting GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { MenuItem, Select, Tooltip } from '@mui/material';
+import { useMemo } from 'react';
+import { useTranslation } from 'react-i18next';
+import { useActiveElements, useElements } from '../../../state';
+import { useLineThickness } from './useLineThickness';
+
+const STROKE_WIDTHS = [4, 6, 8, 10, 12, 14, 16];
+
+export function PolylineThicknessSelect() {
+  const { t } = useTranslation('neoboard');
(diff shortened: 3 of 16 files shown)


packages/react-sdk/src/components/elements/polyline/getRenderProperties.ts
```diff
@@ -26,7 +27,7 @@ export function getRenderProperties(
 ): ElementRenderProperties & PolylineRenderProperties {
   return {
     strokeColor: element.strokeColor,
-    strokeWidth: 4,
+    strokeWidth: element.strokeWidth || defaultStrokeWidth,
```
nurjinjafar, line 30: maybe we want only to exclude null or undefined values like: `strokeWidth: element.strokeWidth ??defaultStrokeWidth, `
or was it on purpose here to default to `defaultStrokeWidth` if the strokeWidth=0?

packages/react-sdk/src/state/crdt/documents/elements.ts
nurjinjafar, line 136: we need to add the: `strokeWidth: Joi.number().strict()` in here as well and since the `strokeWidth` is only applied to polyline and not line then maybe this might be better:   
```
strokeWidth: Joi.number().strict()
    .when('kind', {                                                                                                                                                                        
      is: 'polyline',                                                                                                                                                                      
      then: Joi.optional(),                                                                                                                                                                
      otherwise: Joi.forbidden(),
```

packages/react-sdk/src/components/ElementBar/PolylineThicknessSelect/PolylineThicknessSelect.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const strokeWidths = useMemo(
+    () =>
+      STROKE_WIDTHS.includes(lineThickness)
+        ? STROKE_WIDTHS
+        : [...STROKE_WIDTHS, lineThickness].sort((a, b) => a - b),
+    [lineThickness],
+  );
+
+  const hasPolylinePath = Object.values(elements).find(
+    (element) => element.type === 'path' && element.kind === 'polyline',
```
nurjinjafar: maybe we could reuse the existing function `isPolyline` and `activeElements` ,  `elements` in `useLineThickness()` and move the logic for `hasPolyline` and just use it in here

nurjinjafar:
i think the https://github.com/nordeck/matrix-neoboard/blob/main/docs/model/crdt-documents.md#pathelement document is not updated to include the strokewidth.

packages/react-sdk/src/components/elements/polyline/getRenderProperties.ts
segnord: Yes, zero or null thickness is basically "unset"  and we should will use default.

packages/react-sdk/src/state/crdt/documents/elements.ts
segnord: Ok, I think just limiting it to polyline is too strict, I will make it optional because we have a default when it's unset.

HarHarLinks:
> i think the https://github.com/nordeck/matrix-neoboard/blob/main/docs/model/crdt-documents.md#pathelement document is not updated to include the strokewidth.

I think we need to start getting into a mindset where changes to the document format (e.g. features that require storage) cause a version bump in the document. There needs to be a versioning strategy around this, e.g. major versions for hard incompatibilities (like infinite canvas) and minor versions for features that fall back with degradation on the outdated side, e.g. line thickness not showing up.

segnord:
> > i think the https://github.com/nordeck/matrix-neoboard/blob/main/docs/model/crdt-documents.md#pathelement document is not updated to include the strokewidth.
> 
> I think we need to start getting into a mindset where changes to the document format (e.g. features that require storage) cause a version bump in the document. There needs to be a versioning strategy around this, e.g. major versions for hard incompatibilities (like infinite canvas) and minor versions for features that fall back with degradation on the outdated side, e.g. line thickness not showing up.

In the migrations.ts we say that `Versioning documents should be avoided though and only be used for breaking changes!`

packages/react-sdk/src/components/ElementBar/PolylineThicknessSelect/PolylineThicknessSelect.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+        widgetApi={widgetApi}
+      >
+        <LayoutStateProvider>
+          <Toolbar>{children}</Toolbar>
+        </LayoutStateProvider>
+      </WhiteboardTestingContextProvider>
+    );
+  });
+
+  it('should render without exploding', async () => {
```
maheichyk, line 81: No test here to select a different stroke width value and see it been applied

packages/react-sdk/src/components/ElementBar/PolylineThicknessSelect/PolylineThicknessSelect.tsx
```diff
@@ -0,0 +1,83 @@
...
+    useLineThickness();
+
+  const strokeWidths = useMemo(
+    () =>
+      STROKE_WIDTHS.includes(lineThickness)
+        ? STROKE_WIDTHS
+        : [...STROKE_WIDTHS, lineThickness].sort((a, b) => a - b),
+    [lineThickness],
+  );
+
```
maheichyk, line 36: Why we have this logic to support value in `lineThickness` if it is not supported by known stroke widths? So far looks not relevant to me.

packages/react-sdk/src/components/ElementBar/PolylineThicknessSelect/PolylineThicknessSelect.tsx
```diff
@@ -0,0 +1,83 @@
...
+  if (!firstSelectedPolyline) {
+    return null;
+  }
```
maheichyk, line 39: We should not render this component at all if no polyline is selected

packages/react-sdk/src/components/ElementBar/PolylineThicknessSelect/PolylineThicknessSelect.tsx
```diff
@@ -0,0 +1,83 @@
...
+      }}
+    >
+      {strokeWidths.map((value) => {
+        return (
+          <MenuItem value={value} key={value}>
+            {lineThickness === value && (
+              <Tooltip title={label}>
+                <div>{value}</div>
+              </Tooltip>
+            )}
```
maheichyk, line 75: Why a Tooltip with a div for a selected value only?

packages/react-sdk/src/components/ElementBar/PolylineThicknessSelect/useLineThickness.ts
```diff
@@ -0,0 +1,93 @@
...
+import {
+  Element,
+  ElementUpdate,
+  PathElement,
+  useActiveElements,
+  useElements,
+  useWhiteboardSlideInstance,
+} from '../../../state';
+import { useLayoutState } from '../../Layout';
+import { defaultStrokeWidth } from '../../Whiteboard/constants';
```
maheichyk, line 27: Is not a good idea to import constants from ../../Whiteboard/constants if we have a separate component not coupled to Whiteboard. It gets cluttered very quickly.

packages/react-sdk/src/components/ElementBar/PolylineThicknessSelect/useLineThickness.ts
```diff
@@ -0,0 +1,93 @@
...
+   * Returns the fist polyline in the selection, if exists.
+   */
+  firstSelectedPolyline?: PathElement;
+};
+
+function isPolyline(element: Element): element is PathElement {
+  return element.type === 'path' && element.kind === 'polyline';
+}
+
+export const useLineThickness = (): UseLineThicknessResult => {
```
maheichyk, line 54: Named `useLineThickness` with line and not polyline term which is inconsistent to other changes and confusing as all logic in this PR is about polyline and currently NOT applies to lines.

maheichyk, line 54: Have to have tests, also not tested by PolylineThicknessSelect.test.tsx, especially apply logic

packages/react-sdk/src/components/elements/line/getRenderProperties.ts
```diff
@@ -29,7 +30,7 @@ export function getRenderProperties(
 
   return {
     strokeColor: element.strokeColor,
-    strokeWidth: 4,
+    strokeWidth: defaultStrokeWidth,
```
maheichyk, line 33: Would be nice to have this PR to be applied to lines

packages/react-sdk/src/components/elements/polyline/getRenderProperties.ts
maheichyk, line 30: Maybe still `element.strokeWidth ?? defaultStrokeWidth` ?

Actually zero and null should be considered as invalid values and not pass validaiton

packages/react-sdk/src/components/Layout/useLayoutState.tsx
```diff
@@ -60,6 +60,7 @@ type LayoutState = {
   activeShapeTextShade: number;
   activeStartLineMarker: LineMarker | undefined;
   activeEndLineMarker: LineMarker | undefined;
+  activePolylineThickness: number | undefined;
```
maheichyk, line 63: The use of  term `Thickness` is inconsistent or redundant in the context of PR code changes, you have already introduced `strokeWidth` to `elements` which is much more relevant to the PR changes idea. Let's not use another term in the code. Here and in other places of this PR.

packages/react-sdk/src/components/Layout/useLayoutState.tsx
```diff
@@ -151,6 +153,10 @@ export function LayoutStateProvider({ children }: PropsWithChildren<{}>) {
   };
   const [isRotating, setIsRotating] = useState<boolean>(false);
 
+  const [activePolylineThickness, setActivePolylineThickness] = useState<
+    number | undefined
+  >();
+
```
maheichyk, line 159: Maybe we could put the default value here, at least use it, we do similar for other features, maybe it would simplify things, worth to try.

packages/react-sdk/src/components/Whiteboard/Draft/createShape.ts
```diff
@@ -184,5 +186,9 @@ export function createShapeFromPoints({
     connectedElementEnd,
   };
 
+  if (kind === 'polyline') {
+    pathElement.strokeWidth = strokeWidth;
```
maheichyk, line 190: Sets explicitly undefined when strokeWidth is undefined. Should not set it at all in this case.

packages/react-sdk/src/components/Whiteboard/Draft/DraftLineChild.tsx
```diff
@@ -86,18 +89,19 @@ export const DraftLineChild = ({
       setConnectElementIds([]);
     }
   }, [
-    setActiveTool,
     cursorPoints,
+    setConnectElementIds,
```
maheichyk, line 93: Seems like dependencies are change more then needed

packages/react-sdk/src/components/Whiteboard/constants.ts
```diff
@@ -43,3 +43,5 @@ export const stickyColor = '#ffefc1';
 export const zoomStep = 0.1;
 export const zoomMax = 4;
 export const zoomMin = 0.05;
+
+export const defaultStrokeWidth = 4;
```
maheichyk, line 47: Should have a better place for it. Currently all the usages are outsides of it are outside of `Whiteboard` component.

packages/react-sdk/src/components/Whiteboard/WhiteboardHost.test.tsx
```diff
@@ -1646,4 +1647,73 @@ describe('<WhiteboardHost/>', () => {
     expect(screen.queryByTestId('resize-element')).not.toBeInTheDocument();
     expect(screen.queryByTestId('element-bar-wrapper')).not.toBeInTheDocument();
   });
+
+  it('should change the thickness of a selected polyline without a stroke width', async () => {
```
maheichyk, line 1651: Should not be here, should be in PolylineThicknessSelect.test.tsx

packages/react-sdk/src/state/crdt/documents/elements.ts
```diff
@@ -133,6 +134,7 @@ export const pathElementSchema = elementBaseSchema
     kind: Joi.string().valid('line', 'polyline').required(),
     points: Joi.array().items(pointSchema).required(),
     strokeColor: Joi.string().required(),
+    strokeWidth: Joi.number().optional(),
```
maheichyk, line 137: It is `optional` by default, should have `strict`. Zero, null and other invalid values should be validated.

Should update tests in elements.test.ts

maheichyk (changes requested):

packages/react-sdk/src/components/ElementBar/PolylineThicknessSelect/PolylineThicknessSelect.tsx
segnord: Just a guard that we receive a line thickness from a document that's not in the list so that it's actually displayed

packages/react-sdk/src/components/ElementBar/PolylineThicknessSelect/PolylineThicknessSelect.tsx
segnord: I will double check,  but this should accomplish what you are saying (if it couldn't find even 1 polyline it's falsy)

packages/react-sdk/src/components/ElementBar/PolylineThicknessSelect/PolylineThicknessSelect.tsx
segnord: Otherwise the tooltip showed for items inside the list itself

packages/react-sdk/src/components/ElementBar/PolylineThicknessSelect/useLineThickness.ts
segnord: OK, I will move the constant closer to the lines somewhere

packages/react-sdk/src/components/elements/line/getRenderProperties.ts
segnord: I think would be better as a followup PR

packages/react-sdk/src/components/elements/polyline/getRenderProperties.ts
segnord: This will break backwards compatibility if a document was created by old versions that set this field to  undefined

packages/react-sdk/src/components/Layout/useLayoutState.tsx
segnord: Ok will rename to activeStrokeWidth

packages/react-sdk/src/components/Layout/useLayoutState.tsx
segnord: Ok, we can just have default in the state, so no undefined value

packages/react-sdk/src/state/crdt/documents/elements.ts
segnord: Ok, I think it's better to just remove this rule just to maintain compatibility with old documents

.changeset/true-snails-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+'@nordeck/matrix-neoboard-react-sdk': minor
+---
+
+"Line thickness" control added for selected polyline elements in the Element Bar. The last changed thickness value will be used when adding a new polyline in the current browser session, the default line thickness is 4.
```
maheichyk, line 5: Suggest to use stroke width term, mention polyline as we apply this PR for it only, should make changes text shorter with less details

packages/react-sdk/src/components/ElementBar/PolylineThicknessSelect/useLineThickness.ts
segnord: New place for it will be in `components/common/consts.ts`

packages/react-sdk/src/components/ElementBar/PolylineThicknessSelect/useLineThickness.ts
segnord: New name will be `usePolylineStrokeWidth`

